PR #3, opened 2024-09-17
Add typing

samwaseda:
Originally planned [here](https://github.com/pyiron/elaston/pull/52) but I decided to make it internally a bit less complicated.

```python
@units
def get_speed_onto(
    distance: u(float, "meter"), time: u(float, "second")
) -> u(float, "meter/second"):
    return distance / time
```

The functionality is the same as shown in the previous PRs. What happens internally is:

```python
print(u(float, "meter/second"))
```

Output: `typing.Annotated[float, 'meter/second', None]`

And it's up to the parser to figure out what to do with the additional information. The last item in the list is meant to be used for the ontology type

### DIFF
--- a/tests/unit/test_units.py
+++ b/tests/unit/test_units.py
@@ -9,18 +9,20 @@ from pint import UnitRegistry
 def get_speed_multiple_args(
     distance: u(float, "meter"),
     time: u(float, "second"),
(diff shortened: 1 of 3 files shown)


samwaseda:
@liamhuber Sorry I merged it too fast...

PR #4, opened 2024-09-17
Allow multiple return

samwaseda:
This should be an important feature for workflows.

Taken from the tests:

```python
@units
def get_speed_multiple_outputs(
    distance: u(float, "meter"), time: u(float, "second"), duration: u(float, "second")
) -> (u(float, "meter/second"), u(float, "meter/second")):
    return distance / time, distance / duration
```

The only problem might be that it gets fairly chaotic, but maybe that's something we have to live with...

### DIFF
--- a/tests/unit/test_units.py
+++ b/tests/unit/test_units.py
@@ -2,13 +2,12 @@ import unittest
 from uniton.typing import u
 from uniton.converter import units
 from pint import UnitRegistry
-from typing import Tuple
 
 
 @units
 def get_speed_multiple_outputs(
     distance: u(float, "meter"), time: u(float, "second"), duration: u(float, "second")
-) -> Tuple[u(float, "meter/second"), u(float, "meter/second")]:
+) -> (u(float, "meter/second"), u(float, "meter/second")):
     return distance / time, distance / duration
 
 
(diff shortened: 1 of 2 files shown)


samwaseda:
@liamhuber

liamhuber (approved):
Do you really need to use `Tuple`, or can you just put them in a tuple? Also for readability, I'd suggest putting each element of the return hint on its own line. Otherwise lgtm, and I agree it is an important extension for various real applications.

samwaseda:
> Do you really need to use `Tuple`, or can you just put them in a tuple?

That's actually a brilliant point.

> Also for readability, I'd suggest putting each element of the return hint on its own line.

Yeah but for this one I leave it like this because it's only unit tests.

liamhuber:
> Yeah but for this one I leave it like this because it's only unit tests.

I find the tests are actually a place I go to fairly often in open source in order to gain a deeper understanding of what something is doing -- documentation can get out of date, but the tests reveal what the dev is actually guaranteeing with their code! So from that perspective I've been working to improve the readability of my own tests. This is absolutely fine-grained polish though, and I still think this is a great PR 👍

PR #11, opened 2024-10-15
update github action

samwaseda:


mbruns91:
After some issues with my release (sorry, but I thinks it's ok now) Another issue occurs https://github.com/pyiron/uniton/actions/runs/11348195667/job/31561387155

It seems that the newest ubuntu-runner (??? at least that's my impression) doesn't allow `pip` to be run casually.

so (as suggested in the error output) in the [original `write_docs.yml` action](https://github.com/pyiron/actions/blob/9d25e72e5a4a83d8347165bb8447f8a6dd24fdbf/write-environment/action.yml#L19) it should be changed 
`pip install pyyaml` -> `apt install python3-yaml` (I checked the name of the correct package on [launchpad.net](https://launchpad.net/ubuntu/+source/pyyaml)).

@liamhuber, what do you think?

liamhuber:
> After some issues with my release (sorry, but I think it's ok now) 

🚀 

> another issue occurs https://github.com/pyiron/uniton/actions/runs/11348195667/job/31561387155
>
> It seems that the newest ubuntu-runner (??? at least that's my impression) doesn't allow `pip` to be run casually to protect system-python.

~Can we confirm this? This can be tested by modifying the runner explicitly in the workflow to be an older image.~

I found it via google, no need to confirm: https://github.com/actions/runner-images/issues/10781

> so (as suggested in the error output) in the [original `write_docs.yml` action](https://github.com/pyiron/actions/blob/9d25e72e5a4a83d8347165bb8447f8a6dd24fdbf/write-environment/action.yml#L19) it should be changed `pip install pyyaml` -> `apt install python3-yaml` (I checked the name of the correct package on [launchpad.net](https://launchpad.net/ubuntu/+source/pyyaml)).
> 
> @liamhuber, what do you think?

IMO this is a problem with the image, not a problem with the workflow. In practice we only run this workflow on ubuntu, but in principle this is not specified in the action itself which specifies a `using: "composite"`. Switching from `pip` to `apt` would make the action unusable on windows, yeah?

A better solution would be to manually specify the runner input. Thankfully, this is easily exposed in the reusable workflows we're leveraging. Please try specifying the `runner` input as `"ubuntu-22.04"` to override the default of `"ubuntu-latest"`. 

The downside is that we'll need to do this in each downstream repo, the upside is that there's nothing intrinsically _wrong_ with specifying a particular runner, and we don't wind up messing around with our infrastructure just to accommodate the fact someone made a mistake upstream. I strongly prefer the verbosity over making an unnecessary structural change.

mbruns91:
Ok, so for this repo, I totally agree with you, @liamhuber. For the actions *in gerneral* I think it is worth to consider to go another route. But it's a better idea  - I guess - to have this conversation [here](https://github.com/pyiron/actions/pull/134).

To wrap this one up here (@samwaseda: I can do this tomorrow):
- go back to `actions@v3.3.2`
- specify the runner `runner: "ubuntu-22.04"`

liamhuber:
> go back to `actions@v3.3.2`

`...@actions-3.3.2`

`@v3.3.2` will target the (mutable) branch rather than the release (an immutable tagged commit). In this case, we control the actions repo so realistically there is no security concern, but in principle we want the immutable one for safety.

mbruns91:
I don't know if it's an issue on my side, but I can't push to this repo.
(I just tested to push something to another repo of the pyiron organization and it works fine.)

samwaseda:
now?

mbruns91:
`preview-release.yml` uses `pyiron/actions/.github/workflows/pyproject-release.yml@actions-3.3.2`, so I out-commented the line where a runner is specified.

I think we should fix this upstream in pyiron/actions, i.e. exposing `runner` as optinal input in [`pyiron/actions/.github/workflows/pyproject-release.yml`](https://github.com/pyiron/actions/blob/main/.github/workflows/pyproject-release.yml)

liamhuber:
Super, glad to see this fix worked as expected. @mbruns91, I totally agree the other workflow should also expose the runner as an argument.

PR #25, opened 2024-12-11
Correct action trigger

samwaseda:


niklassiemer:
Since this was after the release, this action was not performed correctly on the release. And I think the pyiron account would be needed to be added to the https://pypi.org/project/semantikon/. Or do you somehow allow github to push to your PyPI.

samwaseda:
So I tried to re-release v0.0.2. That's not enough? And I added `pyiron/semantikon` as a trusted publisher. Not sure if that's what I needed to do...

PR #50, opened 2025-02-10
Move ontology

samwaseda:
In the ongoing efforts to define the interface between semantikon and workflow managers [here](https://github.com/pyiron/pyiron_workflow/issues/576), I started moving the knowledge graph part to semantikon.

### DIFF
--- a/tests/unit/test_ontology.py
+++ b/tests/unit/test_ontology.py
@@ -1,5 +1,6 @@
 import unittest
-from rdflib import Graph, OWL, Namespace, URIRef
+from rdflib import Graph, OWL, Namespace, URIRef, Literal, RDF, RDFS
+from owlrl import DeductiveClosure, OWLRL_Semantics
 from semantikon.typing import u
 from semantikon.converter import (
     parse_input_args,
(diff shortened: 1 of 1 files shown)


samwaseda:
@liamhuber How do I run integration tests?

liamhuber:
It's not like `pyiron_base` or whatever where you need to explicitly add a tag to run them -- they always run, but the compromise we make is that they only run on a single OS-python version pair.

samwaseda:
Oh ok it's part of Test-and-Coverage. I didn't know that. Is there a possibility to use a different yml file for this?

samwaseda:
So the idea is to move the whole thing from `pyiron_ontology` to `semantikon`, but if I understand the current plan correctly, we don't want `semantikon` to depend on `pyiron_workflow`. But I wanted the integration tests to test the `semantikon` - `pyiron_workflow` interactions.

liamhuber:
Ah. Yeah, in the long run I guess I'd rather have those tests in the `pyiron_workflow` repo, but I like your action plan. That definitely sounds like something that is appropriate for the notebooks, as it's a good demo of "and here is the type of cool thing you can do with `semantikon`". So, I'd really not just have the test but annotate the notebook nicely enough that it really does serve the dual purpose of testing and education.

PR #52, opened 2025-02-13
Refactor

samwaseda:


samwaseda:
To be honest I don't really like the current structure either, but since I don't have a better idea right now and there are so many other things to do I leave it like this for now.

PR #59, opened 2025-02-19
Test readme

### LINKED ISSUE #58: :bug: README examples don't run
Docs examples don't run; critically there are no modules `semantikon.convert` nor `semantikon.converters` but rather `semantikon.converter`. This went undetected because the readme examples are not formatted for doctest to read them

liamhuber:
Closes #58 

- Reformatting was trivial, but should later be applied to docstring examples in a similar way.
- Module updates were trivial, just variations on the path name
- Print results were seriously out of date with the example code, but I also just trivially copied and pasted whatever the test was saying I actually got to force the test to pass.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -28,7 +28,7 @@ You can also type-hint the inputs and outputs of a function using a class, i.e.:
 
 ```python
 >>> from semantikon.typing import u
->>> from semantikon.convert import semantikon_class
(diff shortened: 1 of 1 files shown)


samwaseda (approved):
Ah nice thanks for taking care of it. I'm fine as long as the tests pass

PR #115, opened 2025-04-28
Add visualization function

samwaseda:
With this PR, you can visualize knowledge graphs via:

```python
graph = get_knowledge_graph(data)
visualize(graph)
```

It's very primitive, but since many people struggle to visualize knowledge graphs maybe it's better than nothing


samwaseda:
I guess later on we can think about giving different colors to nodes, channels etc.

PR #122, opened 2025-05-03
Use get_type_hints instead of inspect

samwaseda:
`get_type_hints` can deal with future annotations, although forward referencing anyway does not work.

What works:

```python
from __future__ import annotations
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from path_to_my_class import MyClass

def f(x: MyClass):
    return x
```

What does not work:

```python
from __future__ import annotations
from typing import TYPE_CHECKING, Annotated, get_type_hints

if TYPE_CHECKING:
    from path_to_my_class import MyClass

def f(x: Annotated[MyClass, "test"]):
    return x


get_type_hints(f)
```

-> `NameError: name 'MyClass' is not defined`

### DIFF
--- a/semantikon/converter.py
+++ b/semantikon/converter.py
@@ -94,12 +94,14 @@ def parse_output_args(func: callable):
         `label`, `triples`, `uri` and `shape`. See `semantikon.typing.u` for
         more details.
     """
-    sig = inspect.signature(func)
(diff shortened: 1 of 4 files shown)


samwaseda:
I had to implement a whole different function, because `get_type_hints` does not return the metadata of the type hints for python 3.10.

PR #128, opened 2025-05-07
[patch] Update README

samwaseda:
I added some motivation and URI's to the type hints


samwaseda:
The ontology part is not fully implemented (because there is no workflow manager), but I merge it like this for now.

PR #147, opened 2025-05-26
Add mypy

samwaseda:
As the assertion tests in #139 seem a bit redundant, I wanted to see if `mypy` can be applied here as well.


samwaseda:
As we can clearly see, it's gonna be a long way to go, but at the same time it's also a nice side activity when I don't want to use my brain too hard, so I keep this PR open for some time and hopefully `mypy`-compatibility will be achieved at some point.

samwaseda:
I added it in one commit, but it doesn't look like it helped

samwaseda:
Adding `--non-interactive` helped!

PR #148, opened 2025-05-26
[patch] Modify output parser to include the output names

samwaseda:
In the PR (which I haven't opened), I realized that the current output parser, which relies on the output type hints, cannot detect whether the function actually has a return statement or not, so with `get_return_expressions` I check whether there is a return statement or not.


samwaseda:
I just realized that there was a very similar function (`get_return_variables`) which was already defined. Since the new implementation looks somewhat better, I replaced it.

PR #154, opened 2025-05-29
Update pyproject

### LINKED ISSUE #153: pyproject missing dependencies
relative to the `.ci_support/environment.yml`. 

If they're only used for tests that's ok, but if they appear in the source code itself (e.g. `networkx`), they should at least appear as optional dependencies in the toml file.

samwaseda:
Closes #153

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,13 @@ classifiers = [
 ]
 dependencies = [
     "numpy==2.2.3",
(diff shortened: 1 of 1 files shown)


samwaseda:
@liamhuber It looks like there’s a problem with `python-graphviz`, but I don’t quite understand why only this specific package is causing a problem. Could you maybe look into the error?

Issue #162, opened 2025-05-30
Refactoring `_get_nodes`


liamhuber:
I think it would help readability if `_to_node_dict_entry` were complete and returned a full node-dict entry. This would make it easier to draw a connection to a `Function` dataclass. By the same token, it appears to specifically mean a _function_ node dict entry, as the composite workflow entries are handled in the previous clause, so a more specific name would help.

Also, the second `if hasattr(func, "_semantikon_metadata"):` should be redundant -- we can move it's body inside the first clause.

liamhuber:
> Also, the second if hasattr(func, "_semantikon_metadata"): should be redundant -- we can move it's body inside the first clause.

I can't read; the first is "_semantikon_workflow", the second is "_semantikon_metadata". These are not the same words 🤦‍♂

samwaseda:
Historically, `_semantikon_workflow` and `_semantikon_metadata` were attached to the function by simple assignment, i.e.:

```python
def f(x):
    return x

f._semantikon_metadata = some_metadata
f._semantikon_workflow = some_workflow
```

This had to be changed because `mypy` was complaining that an ordinary python function does not have these attributes. But except for being two different attributes, they don't have to belong to different classes, so it might be a good idea to have just one to contain both of them.

liamhuber:
We can independently decorate a function with `@u` and `@workflow`, right? I guess that's actually the case that delivers https://github.com/pyiron/semantikon/pull/164#issue-3104239390

I would be fine with any of
- One class with both attributes
- Two separate classes which preserve class mro when one is applied to a function that is already the other
- An explicit diamond structure with a base class, two child classes each have one attribute, and a third child class that is an explicit child of both the independent ones

I think it's just a matter of experimenting a bit and seeing which implementation feels most robust.

At the end of the day, I think it would be nice to replace the `if hasattr(` tests with `if isinstance(` tests using the resulting classes.

liamhuber:
@samwaseda here is a demo of what I meant in https://github.com/pyiron/semantikon/pull/164#issue-3104239390, which I don't think is covered by https://github.com/pyiron/semantikon/pull/165:

```python
from semantikon.typing import u
from semantikon.workflow import workflow

def add(x, y):
    z = x + y
    return z

@workflow
@u(some_key="a value")
def double_decorated(a, b):
    n = add(a, b)
    return n

[...]

PR #164, opened 2025-05-30
Refactor get nodes

liamhuber:
I was working on #162 and I discovered that the case of a function having both `_semantikon_workflow` and `_semantikon_metadata` at the same time never comes up in the current test suite. @samwaseda, it's not immediately clear to me if this is merely a result of the test suite being insufficient, or if there is a structural reason that these things shouldn't coexist.


samwaseda:
Ah this can totally happen. That's the case when the user decides to give metadata to a workflow.

PR #173, opened 2025-06-02
:fire: HOTFIX reformat README

liamhuber:
My IDE deformatted it, and I sloppily let that through the last PR


samwaseda:
How did it happen XD

liamhuber:
If I knew exactly I might be able to turn it off 🤦‍♂️ I used `Refactor->Move` in PyCharm, and I think it found and moved something in the README example, but then as part of the refactoring on that file it reformats the whole example. In particular splitting `>>>` to `>> >` and breaking lines. 

I should have noticed that the diff was way too many lines, but I was being hasty.

PR #156, opened 2025-05-29
[WIP] Use dataclasses for more explicit data structure

liamhuber:
An example for how we might make progress on #155.

I didn't want to spend too much time on this until I got your feedback on #155, @samwaseda, but I think I got far enough that it lets us make the conversation there more concrete.

PR #174, opened 2025-06-02
Structured metadata

### LINKED ISSUE #159: How to handle annotation flexibility?
Following up on

https://github.com/pyiron/semantikon/issues/155#issuecomment-2922392331 

and 

https://github.com/pyiron/semantikon/issues/155#issuecomment-2922648316

I started to go through and add metadata dataclasses in #156. 

It was easy enough to find the key elements by studying the named kwargs to `semantikon.typing.u`. These differ depending whether we're decorating a function or type, but the latter is a superset of the former:

https://github.com/pyiron/semantikon/blob/e3f084ebd35668846113ee72ae5e8825a716f8da/semantikon/dataclasses.py#L45-L65

For function decoration, it was that simple and immediately worked -- other than updating the default type hints to play well with `_VariadicDataclass` and taking the liberty of re-ordering arguments.

You can see in the copied snippet above, that for type decoration it's a bit more complicated as the tests really exploit the decorator's `**kwargs`. As a stop-gap solution I just added each of these to the dataclass. Obviously this isn't a real solution, but right now I'm just trying to move in the dataclass direction without making _any_ changes to the existing functionality (and thus no changes to the existing tests).

Going to the tests and looking where these fields are used, I can already intuit the sort of downstream use you're hoping to make available (indeed, we've talked about the idea explicitly in the case of `cancel`!) But, of course, these tests are just a skeletal framework to make sure that flexibility is in place, and not a concrete usage.

So,
- Annotation flexibility appears very useful
- Field flexibility is not something dataclasses do
- We like dataclasses

I would thus suggest that we follow your comment that I linked, and add `CoreMetadata.meta: dict[str, Any]`. This can be populated by whatever the user passes to `**kwargs` in the decorator for both the function- and type-decorating uses of `u`. In this way, the dataclass would be rigidly defined but there would still be a place for users to slap whatever extra information they want. The only catch is that they are then also responsible for knowing what it's called and retrieving it from the `meta` dictionary.

Thoughts, @samwaseda?

liamhuber:
Closes #159 

I poached a bunch of dataclass stuff from #156; I'll merge this in there and then stack that on top of this.

In the meantime, this introduces the `CoreMetadata` shared by both functions and types, and a `TypeMetadata` that extends the fields for types. All the extra stuff gets shoved into a `extra: dict[str, Any] | MISSING` field. Right now only types have access to this, not functions, because the current test suite didn't need it for functions, but we could turn it on for them later. For consistence, `parse_metadata` now also returns an actual `TypeMetadata` instead of just a dict in the shape of one. Modifications to the existing tests were basically just adjusting access pathing.

So now,

```python
from semantikon.converter import parse_input_args

@u(uri="abc")
def get_speed(
    distance: u(float, units="meter", something_extra=42, something_else="Not a TypeMetadata field"),
    time: u(float, units="second") = 1.0,
) -> u(float, units="meter/second", label="speed"):
    return distance / time

parse_input_args(get_speed)
```

Gives the unrecognized fields back nested inside the `extra` field. By the time they get to `parse_input_args` they've all been converted back to dictionaries anyhow and we get:

```
{'distance': {'units': 'meter',
  'extra': {'something_extra': 42,
   'something_else': 'Not a TypeMetadata field'},
  'dtype': float},
 'time': {'units': 'second', 'dtype': float, 'default': 1.0}}
```

Note that I kept `**kwargs` available for `u` anyhow, because we don't want to encourage mutable defaults like `u(float, units="meter", extra={"something_extra": 42, "something_else": "Not a TypeMetadata field"})`. The downside I see here is that if a user has a typo in one of the real fields, it will be hard to notice as the value will get shunted off to the `extra` field instead of throwing an error. As @samwaseda [pointed out](https://github.com/pyiron/semantikon/issues/159#issuecomment-2927252711) it also gives a provide/access asymmetry. For the time being, I'm content to simply live with these downsides.

I also added `typeguard` to the dependencies to give us a type-safe way to instantiate the dataclasses from a dictionary.

### DIFF
--- a/semantikon/dataclasses.py
+++ b/semantikon/dataclasses.py
@@ -19,7 +19,7 @@ missing = functools.partial(dataclasses.field, default=MISSING)
 class _HasToDictionary(Iterable[tuple[str, Any]], abc.ABC):
     @abc.abstractmethod
     def __iter__(self) -> Iterator[tuple[str, Any]]:
-        ...
+        pass
 
     def to_dictionary(self) -> dict[str, Any]:
         d = {}


semantikon/dataclasses.py
```diff
@@ -0,0 +1,77 @@
...
+    | RestrictionClause  # Short-hand for a single-clause restriction
+)
+ShapeType: TypeAlias = tuple[int, ...]
+
+
+@dataclasses.dataclass(slots=True)
+class CoreMetadata(_VariadicDataclass):
+    uri: str | Missing = missing()
+    triples: TriplesLike | Missing = missing()
+    restrictions: RestrictionLike | Missing = missing()
```
samwaseda, line 69: Was `restrictions` also in functions? I don't think we can apply it right now, so we could also move it to `TypeMetadata`.

samwaseda: ok now I see that it was there. Maybe we can move it in a different PR since it's totally unrelated.

semantikon/dataclasses.py
```diff
@@ -0,0 +1,77 @@
...
+        return "<MISSING>"
+
+
+MISSING = Missing()
+missing = functools.partial(dataclasses.field, default=MISSING)
+
+
+class _HasToDictionary(Iterable[tuple[str, Any]], abc.ABC):
+    @abc.abstractmethod
+    def __iter__(self) -> Iterator[tuple[str, Any]]: ...
```
samwaseda, line 21:
```suggestion
    def __iter__(self) -> Iterator[tuple[str, Any]]:
        ...
```

samwaseda: Codacy was complaining

PR #185, opened 2025-06-10
Service config

liamhuber:
Update to the latest pyiron/actions and switch to codecov

@samwaseda, I think this may be why we're still getting coveralls. I recently did this stuff somewhere else (pyiron_ontology?) and got mixed up that it wasn't done here -- my bad!

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -8,5 +8,5 @@ on:
 
 jobs:
   codeql:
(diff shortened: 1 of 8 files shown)


liamhuber:
Ok, well, this is still a good PR, but it doesn't solve the problem. The push-pull workflow is not triggering codacy but it's still happening.

PR #191, opened 2025-06-12
Drop python 3.9 from pyproject

### LINKED ISSUE #188: 3.9 is listed as supported, but not tested
Either we should go back and add it to the test suite, or drop formal support. I'd lean towards dropping it -- 3.9 is done its lifetime by the end of the year anyhow

samwaseda:
Closes #188

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -10,6 +10,7 @@ dependencies:
 - networkx =3.4.2
 - pyshacl =0.30.1
 - graphviz =12.2.1
-- python >=3.10,<3.13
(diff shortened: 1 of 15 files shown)


liamhuber:
I'm not sure if it is necessary or helpful for propagating through to conda forge, but it might be nice to pin the python range in the env yaml too

samwaseda:
> I'm not sure if it is necessary or helpful for propagating through to conda forge, but it might be nice to pin the python range in the env yaml too

hmmm I'm not sure if I know which file you are referring to

PR #194, opened 2025-06-16
Shift upper limit to <3.14

liamhuber:
and test 3.13


liamhuber:
Ubuntu tests for 3.13 passed breezily. I guess there's no reason to not extend support to 3.13?

samwaseda (approved):
Beautiful!

PR #195, opened 2025-06-17
Use textwrap to be able to parse nested functions

samwaseda:
There's not much of a reason to parse local functions (e.g. functions inside another function), but there's also no reason to forbid it, so I included `textwrap` to get around it.

### DIFF
--- a/semantikon/workflow.py
+++ b/semantikon/workflow.py
@@ -2,6 +2,7 @@ import ast
 import builtins
 import copy
 import inspect
+import textwrap
 from collections import deque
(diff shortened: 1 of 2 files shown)


liamhuber:
Local functions are not importable, which is a huge pain. The dedent here is still fine, as I imagine it is also needed for parsing functions defined inside classes -- again, not strict necessary but why not; one can use classes as a sort of pseudo namespace sometimes.

We might need to manually add a check that the things being parsed are importable at some point though?

Issue #146, opened 2025-05-25
How to represent `if` - `elif` - `else`


samwaseda:
Now we come to the final part of our trilogy. Let's take the following example:

```python
def my_condition_one(x, y):
    return x > y

def my_condition_two(x):
    return x > 0

def add(x, y):
    return x + y

def multiply(x, y):
    return x * y

def my_workflow(x, y):
    if my_condition_one(x, y):
        result = add(x, y)
    elif my_condition_two(x):
        z = multiply(x, y)
        result = add(x, z)
    else:
        z = add(x, y)
        result = multiply(y, z)
    return result
```

Following the structure of `while` (cf. #143), the three blocks are wrapped in independent injected nodes:

- if-node
  - test
  - body
- elif-node
  - test
  - body
- else-node
  - body

(body stands for inputs, outputs, nodes and edges)

That means the whole thing would look like this:

```python
{
    "inputs": {"x": {}, "y": {}},
    "outputs": {"result": {}},
    "nodes": {
        "injected_if_0": {
            "test": {
                "function": my_condition_one(x, y),
                "inputs": {"x": {}, "y": {}},
                "outputs": {"output": {}},
            },
            "inputs": {"y": {}, "x": {}},
            "outputs": {"result": {}},
            "nodes": {
                "add_0": {
                    "function": add(x, y),
                    "inputs": {"x": {}, "y": {}},
                    "outputs": {"output": {}},
                }
            },
            "data_edges": [
                ("inputs.x", "test.inputs.x"),
                ("inputs.x", "add_0.inputs.x"),
                ("inputs.y", "test.inputs.y"),
                ("inputs.y", "add_0.inputs.y"),
                ("add_0.outputs.output", "outputs.result"),
            ],
            "label": "injected_if_0",
        },
        "injected_elif_0": {
            "test": {
                "function": my_condition_two,
                "inputs": {"x": {}},
                "outputs": {"output": {}},
            },
            "inputs": {"y": {}, "x": {}, "z": {}},
            "outputs": {"result": {}, "z": {}},
            "nodes": {
                "multiply_0": {
                    "function": multiply,
                    "inputs": {"x": {}, "y": {}},
                    "outputs": {"output": {}},
                },
                "add_0": {
                    "function": add,
                    "inputs": {"x": {}, "y": {}},
                    "outputs": {"output": {}},
                },
            },
            "data_edges": [
                ("inputs.x", "test.inputs.x"),
[...]

PR #200, opened 2025-06-17
Implement if else statement

samwaseda:
Following [this discussion](https://github.com/pyiron/semantikon/issues/146). So far I managed to implement `if` and `else`, and I think one `elif` can also be straightforwardly implemented. More on the [issue page](https://github.com/pyiron/semantikon/issues/146) for the discussion.

### DIFF
--- a/semantikon/workflow.py
+++ b/semantikon/workflow.py
@@ -81,6 +81,21 @@ def _hash_function(func):
     return f"{func.__name__}_{sha256(inspect.getsource(func).encode()).hexdigest()}"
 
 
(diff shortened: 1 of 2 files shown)


samwaseda:
I have no idea how reliably the algorithm works, but the tests passed...

samwaseda:
I don't think `elif` works in this version, but I'm just too tired to think about it, so I merge it.

samwaseda:
I don't think `elif` works in this version, but I'm just too tired to think about it, so I merge it.

PR #211, opened 2025-06-28
[patch] Correct README

samwaseda:
Apparently I should have looked at the errors from #210 more seriously. Hopefully this PR is going to fix them.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -172,7 +172,7 @@ Based on the type hints, `semantikon` can create a knowledge graph of the functi
 
 This creates an `rdflib`-graph, which you can visualize with the programme of your choice. If you use `semantikon.visualize.visualize`,  you can get the following figure:
 
-<img src="../images/knowledge_graph.png" alt="Knowledge Graph Example" width="600"/>
+<img src="../images/knowledge_graph.png" alt="Knowledge Graph Example"/>
 
 
 #### Check node compatibility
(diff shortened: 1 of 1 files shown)


samwaseda:
```
File "/home/runner/work/semantikon/semantikon/tests/integration/../../docs/README.md", line 255, in README.md
Failed example:
    print(validate_values(graph))
Exception raised:
    Traceback (most recent call last):
      File "/home/runner/work/semantikon/semantikon/cached-miniforge/my-env/lib/python3.12/doctest.py", line 1368, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest README.md[49]>", line 1, in <module>
        print(validate_values(graph))
                              ^^^^^
    NameError: name 'graph' is not defined
```

@liamhuber I don't really understand why it's failing, because `graph` is clearly defined in the line before...

samwaseda:
ok I guess the problem comes from the line before, but I don't really understand this error either...

samwaseda:
What's failing is quite obviously `inspect.getmodule`. According to [their documentation](https://docs.python.org/3/library/inspect.html#inspect.getmodule), it returns `None` if the module cannot be determined. I'm not really sure why though.

liamhuber:
I have low expectations for finding a better solution for the readme. I did think of one more workaround though: mention this capability without example in the readme, and move the example over to a tutorial notebook. This changes the presentation, obviously, but immediately resolves the technical problem.

PR #212, opened 2025-06-28
Force doctest to know the namespace

liamhuber:
Ok @samwaseda, here's a solution. The core idea is just to force-assign all the relevant function declarations to some known namespace, then `inspect` and the rest of the tools can find everything OK. I made it as un-invasive as I could by packaging it all as a decorator, but it's still kind of ugly.

Please particularly note that I meaningfully altered your demo! I made the final printout all one line, which is nothing, but I also had to alter "dye_0" to "sell_0". I think this is correct, since "dye_0" can't be known at this time as it's not declared in the `my_wrong_workflow`. Rather, it now reads to me as complaining that the input to `sell` is incorrect, which I agree with. Please double-check that this still communicates what you want.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -147,21 +147,37 @@ Regardless of whether type hints are provided, the interpreter acts only when th
 
 Based on the type hints, `semantikon` can create a knowledge graph of the function. The knowledge graph is a directed acyclic graph (DAG) that contains the inputs and outputs of the function, as well as the units and ontological types.
 
+Please bear with us for a moment while we do some book-keeping to manage our automated README test
+
+```python
+>>> import types, sys
+>>> doctestmod = types.ModuleType("justfordoctests")
+>>> sys.modules["justfordoctests"] = doctestmod
+>>> def register_for_doctest(func):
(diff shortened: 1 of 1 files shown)


liamhuber:
Ah, and I boldly assigned you as reviewer and made these claims at PR-open-time because "it's working on my machine". Moment of truth.....

samwaseda:
Hui that’s a pretty hard constraint. There’s no workaround without touching README?

liamhuber:
I didn't find one with Google or ChatGPT, but I also didn't spend super long searching.

liamhuber:
I mean, when you run it inside doctest the module attribute is just not there; that is difficult to overcome.

You could go into the code and force-assign stuff to main conditionally when inspect can't find the module, but IMO that is a much worse idea than some readme mess.

The last workaround I can think is to put the readme functions somewhere real and just import them here and print their source. That is much safer than an extra condition in the source code, but I don't know if it's really any prettier than this function.

samwaseda (approved):
ok let's merge this one for now to `bugfix`, because I don't think it makes much sense to keep it failing.

PR #214, opened 2025-06-30
Update notebooks

samwaseda:
Following [@liamhuber's suggestion](https://github.com/pyiron/semantikon/pull/211#issuecomment-3016695520) I updated notebooks to include all ontology examples. For now README is untouched.


liamhuber (approved):
LGTM. I appreciate the `pyiron_workflow` shoutout! I'm really optimistic about this integration, both in terms of it being something we can accomplish and something useful. This is a super cool feature/power you're cooking with here.

I found the pedagogy to be good and clear, but here's a few nits if you want them:

## subject replacement

"The following notations are equivalent: ... `"outputs.clothes"`" -- I guess this is a direct reference to the function's return variable. Would it be interesting to later show how this equivalence feature and by-name access works for a function with multiple returns? Might be beyond-scope for this notebook, but it was a question I was left with as a curious reader

## Formatting

I recommend breaking some of the longer lines up in the function definition, e.g.

```python
def get_speed(distance: u(float, uri=EX.distance, units="meter"), time: u(float, uri=EX.time, units="second")) -> u(float, uri=EX.velocity, units="meter/second"):
```

To

```python
def get_speed(
    distance: u(float, uri=EX.distance, units="meter"), 
    time: u(float, uri=EX.time, units="second"),
) -> u(float, uri=EX.velocity, units="meter/second"):
```

And

```python
def wash(clothes: Clothes) -> u(Clothes, triples=((SNS.inheritsPropertiesFrom, "inputs.clothes"), ("outputs.clothes", EX.hasProperty, EX.cleaned))):
```

To

```python
def wash(clothes: Clothes) -> u(
    Clothes, 
    triples=(
        (SNS.inheritsPropertiesFrom, "inputs.clothes"), 
        ("outputs.clothes", EX.hasProperty, EX.cleaned)
    )
):
```

For that matter, a couple of the `restrictions=` lines would similarly benefit.

## Flowchart rendering?

The `flowchart LR` boxes are not rendering for me, it's just flowchart-defining text. Maybe this is intentional?


## Anticipate the linter

Sometimes the linters will complain about re-use of variables. Here I see that we could get more clarity from defining a new variable in a few places, e.g.

```python
def my_correct_workflow(clothes: Clothes) -> int:
    clothes = dye(clothes)
    clothes = wash(clothes)
    money = sell(clothes)
    return money
```

To

```python
def my_correct_workflow(clothes: Clothes) -> int:
    died_clothes = dye(clothes)
    cleaned_clothes = wash(died_clothes)
    money = sell(cleaned_clothes)
    return money
```

(maybe this is counter-productive for your wrong-order demo...but actually shows why the linters like this, as the more verbose variable names make it easier to spot if I'm accidentally doing stuff in a different order than I mean!)

PR #218, opened 2025-07-04
Introduce type checking

### LINKED ISSUE #216: Introduce type checking
So far, only OWL restrictions are available in `semantikon` (which are more difficult than type checking...), but since the NFDI conference is coming up, and generally it should be part of what the reasoner should be able to do, I'm gonna implement it here now. In principle I need [`type_hint_is_as_or_more_specific_than`](https://github.com/pyiron/pyiron_workflow/blob/24fd88769374f20e2edf6a8f209bb2fb716d214f/pyiron_workflow/type_hinting.py#L62) from `pyiron_workflow`, but on the knowledge graph level, which should be easier, since after running a reasoner all the RDFS subclasses should be attached to the object explicitly.

samwaseda:
Closes #216


samwaseda:
@liamhuber Since I need this next week and I know you are busy, I don't ask for your review before merging it, but feel free to leave a comment.

PR #228, opened 2025-07-15
Parse dictionary output

samwaseda:
I started feeling that we should give the possibility of returning `dict` and data classes, simply because it's the de facto standard for input and output parsers.

Before (this works just as well):

```python
def get_output(job_instance):
    ...
    return positions, forces, energy
```

After:

```python
def get_output(job_instance):
    ...
    return {"positions", x, "forces", f, "energy": E}
```

And these two should have the same output keys. This would imply that we would allow something like this:

```python
def my_workflow(some_inputs):
    ...
    output_dict = get_output(job)
    my_input = get_input_of_another_job(energy=output_dict["energy"])
    ...
```

In this case the parsing will be slightly more complicated, but I guess it's still a reasonable amount of work. What do you think of the overall idea? @jan-janssen @liamhuber

### DIFF
--- a/semantikon/converter.py
+++ b/semantikon/converter.py
@@ -172,11 +172,8 @@ def _to_tag(item: Any, count=None, must_be_named: bool = False) -> str:
         return f"output_{count}"
 
 
-def get_return_expressions(
-    func: Callable, separate_tuple: bool = True, strict: bool = False
-) -> str | tuple[str, ...] | None:
-    source = inspect.getsource(func)
-    source = textwrap.dedent(source)
+def _get_return_list(func: Callable, strict: bool = False) -> list[str | tuple[str, ...]]:
+    source = textwrap.dedent(inspect.getsource(func))
     parsed = ast.parse(source)
 
     func_node = next(n for n in parsed.body if isinstance(n, ast.FunctionDef))
@@ -199,7 +196,13 @@ def get_return_expressions(
                 )
             else:
                 ret_list.append(_to_tag(value, must_be_named=strict))
+    return ret_list
 
+
+def get_return_expressions(
+    func: Callable, separate_tuple: bool = True, strict: bool = False
+) -> str | tuple[str, ...] | None:
+    ret_list = _get_return_list(func, strict=strict)
     if len(ret_list) == 0 and not strict:
         return None
     elif len(set(ret_list)) == 1 and (
--- a/semantikon/converter.py
+++ b/semantikon/converter.py
@@ -4,6 +4,7 @@
 
 import ast
 import inspect
+import keyword
 import re
 import sys
 import textwrap
@@ -157,10 +158,17 @@ def _resolve_annotation(annotation, func_globals=None):
         return Annotated[undefined_name, args[1]]
 
 
+def _is_valid_variable_name(name: str) -> bool:
+    return name.isidentifier() and not keyword.iskeyword(name)
+
+
 def _to_tag(item: Any, count=None, must_be_named: bool = False) -> str:
     if isinstance(item, ast.Name):
         return item.id
-    elif must_be_named:
+    elif isinstance(item, ast.Constant):
+        if isinstance(item.value, str) and _is_valid_variable_name(item.value):
+            return item.value
+    if must_be_named:
         raise NotAstNameError(
             "With `must_be_named=True`, item must be captured in an `ast.Name` "
             "variables, i.e only simple variable(-s) not containing any operation or "
@@ -172,7 +180,9 @@ def _to_tag(item: Any, count=None, must_be_named: bool = False) -> str:
         return f"output_{count}"
 
 
-def _get_return_list(func: Callable, strict: bool = False) -> list[str | tuple[str, ...]]:
+def _get_return_list(
+    func: Callable, strict: bool = False
+) -> list[str | tuple[str, ...]]:
     source = textwrap.dedent(inspect.getsource(func))
     parsed = ast.parse(source)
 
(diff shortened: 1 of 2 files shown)


samwaseda:
> I'm a bit confused -- what is a "job" here?

Maybe I should have called it `output_file` or something. It doesn't really matter as long as the function returns a set of values.

liamhuber:
> Maybe I should have called it `output_file` or something. It doesn't really matter as long as the function returns a set of values.

Ah, ok.

So do I understand correctly that you really want

```python
def my_workflow(some_inputs):
    ...
    output_dict = get_output(job)
    my_input = get_input_of_another_job(energy=output_dict["energy"])
    ...
```

To be the right way to write things, regardless of whether `output_dict` uses your first or second definition? If so, I really dislike that. My rationale is that currently the workflows are also just valid python; but here the definition of `my_workflow` above is _only_ compatible with the dictionary-returning definition of `output_dict`.

liamhuber:
Changing tack a little bit, I do rather like the idea of an `@node` parsing returned dictionaries into individual output channels. I was at first a little nervous because I was like "but what if the user is _trying_ to return an entire coherent dictionary?!" but then I realized `pyiron_workflow` has the same problem with tuples, and we already get around it by saying "if you return a tuple but your node decorator has a single output label, we treat the entire tuple as a single return". Of course we could do exactly that same thing with returned dictionaries!

So I actually rather like that these changes to `get_return_expressions` exist. In the medium run, I'd like for `pyiron_workflow` to exploit the function parsing capabilities here, and so being able to parse tuples to output channels _and_ dictionaries to output channels would be super. I think also it would trim down some of the intermediate steps that were necessary to make `pyiron_workflow` interoperable with the Python Workflow Definition https://github.com/pythonworkflow/python-workflow-definition/pull/111

I.e. I like the power being introduced here, I'm just deeply unconvinced of the proposed plan for exploiting it in `@semantikon.workflow.workflow` functions.

liamhuber:
> I.e. I like the power being introduced here, I'm just deeply unconvinced of the proposed plan for exploiting it in `@semantikon.workflow.workflow` functions.

To this end, @samwaseda it may be helpful to me if there were some working example of your intent? The test suite extension here is useful, but doesn't touch the workflow recipe generation.

I checked the branch out and tried things for myself:

```python

def outputs_dict(x, y):
    return {"x": x, "y": y}

def outputs_tuple(i, j):
    return i, j

def add(obj, other):
    sum = obj + other
    return sum

# @workflow
# def my_macro(x, y, i, j):
#     out_dict = outputs_dict(x, y)
#     out_i, out_j = outputs_tuple(i, j)
#     use_it = add(out_dict["x"], out_i)
#     return use_it

@workflow
def my_macro(x, y, i, j):
    out_dict = outputs_dict(x, y)
    out_i, out_j = outputs_tuple(i, j)
    return out_dict, out_i, out_j

my_macro._semantikon_workflow
```

Since the new functionality is not yet propagated to the workflows, on both main and this branch I get the entry

```python
  'outputs_dict_0': {'inputs': {'x': {}, 'y': {}},
   'outputs': {'output': {}},
   'function': <function __main__.outputs_dict(x, y)>,
   'type': 'Function'}},
```

I guess the idea here is that this should have `'outputs': {'x': {}, 'y': {}}`? That's something I can get behind.

From there, I would also be supportive of something like the commented-out workflow, where the tuple-based return and dict-based return both work but the syntax for leveraging them is different. However right now this unsurprisingly gives `NotImplementedError: Only variable inputs supported, got: {'_type': 'Subscript', 'value': {'_type': 'Name', 'id': 'out_dict', 'ctx': {'_type': 'Load'}}, 'slice': {'_type': 'Constant', 'value': 'x', 'kind': None}, 'ctx': {'_type': 'Load'}}`.

If it _did_ work, I would imagine to see

```python
 'edges': [
   ...
   ('outputs_dict_0.outputs.x', 'add_0.inputs.obj'),
   ('outputs_tuple_0.outputs.i', 'add_0.inputs.other'),
   ...
```

Now this is a little bit interesting, because it _demands_ that WfMS interpret both tuples and dictionaries as independent outputs when parsing plain functions. I'm OK with this demand, and would like to support it in `pyiron_workflow`, I just think it's important to clearly note what requirements we're making.

samwaseda:
Ah ok now I understood the confusion. Sorry for `tuple` the behavior is the same as before, meaning with

```python
def get_output(job_instance):
    ...
    return positions, forces, energy
```

you would have to do

```python
def my_workflow(some_inputs):
    ...
    positions, forces, energy = get_output(job)
    my_input = get_input_of_another_job(energy=energy)
    ...
```

samwaseda:
> Now this is a little bit interesting, because it demands that WfMS interpret both tuples and dictionaries as independent outputs when parsing plain functions.

Yeah the more I think about it, the more I feel it's potentially a super dangerous step, because currently the number of outputs can be specified by the number of variables assigned, i.e.:

```python
def f(x, y):
    return x, y

def my_workflow_1(x, y):
    x, y = f(x, y)
    return x, y

def my_workflow_2(x, y):
    z = f(x, y)
    return z
```

So in the first case there are two outputs, and in the second case one. But now if we use `dict`, the output is the same. We can see only from the subsequent lines whether the dictionary is resolved or not, or in the worst case we might have something like:

```python
def f(x, y):
    return {"x": x, "y": y}

def my_workflow(x, y):
    d = f(x, y)
    a = g(d["x"])
    b = h(d)
    ...
```

In this case both multiple outputs and single outputs must be simultaneously dealt with. I don't think such a case can be represented straightforwardly...

PR #231, opened 2025-07-16
Function to issue warning if default value is used

### LINKED ISSUE #227: Check whether default was changed
While creating the post about how to write workflows, I realized that with workflow managers, we can check whether a value was set or not, i.e. these two cases are not the same[^1]:

```python
def f(x):
    return x

f(x=100)
```

and

```python
def f(x=100):
    return x

f()
```

Now, why is this important? That's because there are some parameters, which **should** be changed by the user in the production mode, such as DFT convergence parameters, or the temperature value of MD calculations, even if they happen to be the same as the default parameters. For example, the user might happen to want to run an MD calculation at 300 K, which is `pyiron_atomistics`'s default value, but we could issue a warning if `run()` is called instead of `run(temperature=300)`.

I guess the most natural place to indicate whether the value should be explicitly chosen is the semantikon type hint, such as:

```python
def f(x: u(float, should_be_explicit=True) = 100):
    return x
```

I'm very much aware of the fact that `should_be_explicit` sounds super clumsy, but I was wondering if a feature like this could be useful, or maybe we could think of more cases like this. Thoughts?

[^1]: I guess with @jan-janssen's PWD they are the same

samwaseda:
Closes #227 

While developing the [DAMASK example](https://github.com/samwaseda/damask_local), I strongly felt the need to warn the user if the default values are used, because some of them are not intuitively clear. With this PR, you can write a function like this:

```python
@with_explicit_defaults
def f(x=use_default(3)):
    return x
```

This is largely equivalent to:

```python
def f(x=3):
    return x
```

The only difference is that if the user runs `f()`, it issues a warning, but if it's explicitly set by `f(3)` or `f(x=3)` (or any other value), then it doesn't say anything. You can also find the actual use [in this notebook](https://github.com/samwaseda/damask_local/blob/main/notebooks/tensile_test.ipynb)

### DIFF
--- a/semantikon/converter.py
+++ b/semantikon/converter.py
@@ -462,7 +462,29 @@ def semantikon_class(cls: type) -> type:
     return cls
 
 
-def with_explicit_defaults(func):
+def with_explicit_defaults(func: Callable) -> Callable:
+    """
+    Decorator to marks a value as an explicit default, which can be used to
+    indicate that a value should be replaced with a default value in the
+    context of serialization or processing.
+
+    Args:
+        func: function to be decorated
+
+    Returns:
+        decorated function that replaces explicit defaults with the actual default
+        value and issues a warning if the default is used.
+
+    Example:
+
+    >>> @with_explicit_defaults
+    >>> def f(x=use_default(3)):
+    ...     return x
+
+    >>> f()  # This will return 3, and a warning will be issued.
+
+    >>> f(3)  # This will also return 3 but without any warning.
+    """
     sig = inspect.signature(func)
 
     @wraps(func)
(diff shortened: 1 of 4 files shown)


samwaseda:
I'm asking you guys (@jan-janssen @liamhuber ) for code review, but I'd rather like to know your opinion, and not so much about the code.

liamhuber (approved):
I know you two have been talking about this, so my ignorance here is wilful at this point :joy: but I don't really understand why I would care to do this. But given that I want to, the interface to do so here seems pretty reasonable and intuitive.

By default `warnings.warn` are going to `sys.stderr` or something like that. If I have a logger, I can activate that particular logger catching _all_ warnings. Still, it might be useful to extend `with_explicit_defaults` with an optional logger method so I can choose to always route these warnings in particular there.

PR #233, opened 2025-07-30
Add version to metadata

samwaseda:
I saw there's a possibility to add the version number in [pyiron_database](https://github.com/pyiron/pyiron_database/blob/main/pyiron_database/instance_database/node.py), so I thought maybe it's a good idea to add the field in the function decorator. The idea is maybe something like this:

```python
@u(version="1.5.2")
def f(x):
    ...
```

### DIFF
--- a/semantikon/datastructure.py
+++ b/semantikon/datastructure.py
@@ -68,6 +68,7 @@ class CoreMetadata(_VariadicDataclass):
     uri: str | Missing = missing()
     triples: TriplesLike | Missing = missing()
     restrictions: RestrictionLike | Missing = missing()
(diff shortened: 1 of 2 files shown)


samwaseda:
> I'm always nervous about version numbers that users/devs need to update manually are doomed to fall out of accuracy

Yeah that was the reason why I asked for review. I just have no idea where it should come from, since we don't have any specific location yet. I'm gonna keep this PR open for now then.

samwaseda:
I realized that `pyiron_database` takes `function.__version__`, which makes more senes to me, so I close this PR.

PR #241, opened 2025-08-22
Inherit types with `derive_from`

### LINKED ISSUE #240: ontological annotation `derived_from` does not propagate `type`
Modifying the clothing example slightly and pulling the `owlrl.DeductiveClosure` operation out of `semantikon.ontology.validate_values` for good measure, we see

```python
import owlrl
import rdflib

class Clothes:
    pass

def machine_wash(
    clothes: u(
        Clothes,
        uri=EX.Garmant,
        triples=(EX.hasProperty, EX.MachineWashable),
    )
) -> u(
    Clothes,
    triples=(EX.hasProperty, EX.cleaned),
    derived_from="inputs.clothes",
):
    ...
    return clothes

@workflow.workflow
def wrap_in_wf():
    c = machine_wash()
    return c

recipe = wrap_in_wf._semantikon_workflow

g = onto.get_knowledge_graph(recipe)
owlrl.DeductiveClosure(owlrl.OWLRL_Semantics).expand(g)
for subject in g.subjects(predicate=rdflib.RDF.type, object=EX.Garmant):
    print(subject)
>>> wrap_in_wf.machine_wash_0.inputs.clothes

```

that even though the output of `machine_wash` has `"derived_from"` our input field, it has not inherited the type from the input `uri` field and thus has no `type`.

The remaining behaviour appears to be working fine, IMO. The output nicely inherits the `MachineWashable` information:

```python
for subject in g.subjects(predicate=EX.hasProperty, object=EX.MachineWashable):
    print(subject)
>>> wrap_in_wf.machine_wash_0.inputs.clothes
>>> wrap_in_wf.machine_wash_0.outputs.clothes
>>> wrap_in_wf.outputs.c

```

And the `triple` command for the output is being parsed fine:

```python
for subject in g.subjects(predicate=EX.hasProperty, object=EX.Cleaned):
    print(subject)
>>> wrap_in_wf.machine_wash_0.outputs.clothes
>>> wrap_in_wf.outputs.c

```

If an output field is `derived_from`, my best understanding is that its `rdflib.RDF.type` should be non-empty and congruent with the input type, but the current attack leveraging `SNS.inheritsPropertiesFrom`

https://github.com/pyiron/semantikon/blob/8539e67c324724c42e07499d02d92731580f3881/semantikon/ontology.py#L103-L111

is not doing the trick.

This causes trouble if I then have a downstream node expecting a `clothes: u(Clothes, uri=EX.Garmant)`, as it will fail connection checks.

> If an output field is `derived_from`, my best understanding is that its `rdflib.RDF.type` should be non-empty and congruent with the input type

The trouble is, I'm not sure what "congruent" should mean. If the derived field has no associated `rdflib.RDF.type` (i.e. no `uri=` assigned? Are there other ways we might wind up with a type? I guess users could always assign one manually in `triples=`...), then applying all the types held by the derived-from field seems fine. But what about the other cases?

Derived field specifies...
- No types: safely take types from source
- Less specific types (i.e. superclass(es)): Do nothing?
- More specific types (i.e. subclass(es)): Allow this type narrowing by ignoring source superclasses??
- Completely unrelated types: Fail hard?? How to check? Just let it pass and take the union of all types???

As a starting point, I guess it would be safe enough to search for existing types, and if there are none then inherit from the source, and if there are any to fail hard immediately. We could then work out more subtlety for the other cases going forward...

@samwaseda, I'm happy to write that first rough attack PR, so I don't want to assign you, but I definitely want your thoughts here.

liamhuber:
The actual change is just a one-liner so `derive_from` also snags the  `rdflib.RDF.type` whence the field derives.

Closes #240 

Example:

```python
import rdflib

class Clothes: ...

def machine_wash(
    clothes: u(
        Clothes,
        uri=EX.Garment,
        triples=(EX.hasProperty, EX.MachineWashable),
    )
) -> u(
    Clothes,
    uri=EX.SomethingElse,
    triples=(EX.hasProperty, EX.Cleaned),
    derived_from="inputs.clothes",
):
    ...
    return clothes

@workflow.workflow
def wrap_in_wf(
    clothes: u(
        Clothes,
        uri=EX.Garment,
        triples=(EX.hasProperty, EX.MachineWashable),
    )
):
    washed = machine_wash(clothes)
    return washed
```

Then 

```python
g = onto.get_knowledge_graph(wf_dict=recipe)
onto.validate_values(g)
```

Is non-empty, because the output of `machine_wash` winds up with two completely unrelated types.

```python
g0 = rdflib.Graph()
g0.add((EX.SomethingElse, rdflib.RDFS.subClassOf, EX.Garment))

g = onto.get_knowledge_graph(
    wf_dict=recipe,
    graph=g0,
)
onto.validate_values(g)
```

Is non-empty, because the downstream (output) type is _narrower_ than the upstream (input) type.

```python
g0 = rdflib.Graph()
g0.add((EX.Garmant, rdflib.RDFS.subClassOf, EX.SomethingElse))

g = onto.get_knowledge_graph(
    wf_dict=recipe,
    graph=g0,
)
onto.validate_values(g)
```

Finally passes, since the downstream type is now _broader_ than the upstream type, and thus the upstream is always fulfilling the downstream requirement.

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -248,7 +248,6 @@ def _inherit_properties(
         "    FILTER(?p != rdfs:label)",
         "    FILTER(?p != rdf:value)",
         "    FILTER(?p != ns:hasValue)",
-        "    FILTER(?p != rdf:type)",
         "    FILTER(?p != owl:sameAs)",
         "}",
     )
@@ -306,6 +305,21 @@ def _check_connections(graph: Graph, strict_typing: bool = False) -> list:
     return incompatible_types
 
 
+def _check_types(graph: Graph) -> list:
+    query = f"""
+    PREFIX rdfs: <{RDFS}>
+    PREFIX owl: <{OWL}>
+    
+    SELECT ?subject ?class1 ?class2 WHERE {{
+        ?subject rdfs:subClassOf ?class1 .
+        ?subject rdfs:subClassOf ?class2 .
+        ?class1 owl:disjointWith ?class2 .
+        FILTER(?class1 != ?class2)
+    }}
+    """
+    return list(graph.query(query))
+
+
 def validate_values(
     graph: Graph, run_reasoner: bool = True, strict_typing: bool = False
 ) -> dict[str, list]:
@@ -327,6 +341,7 @@ def validate_values(
         "incompatible_connections": _check_connections(
             graph, strict_typing=strict_typing
         ),
+        "inconsistent_types": _check_types(graph),
     }
 
 
--- a/tests/unit/test_ontology.py
+++ b/tests/unit/test_ontology.py
@@ -221,6 +221,37 @@ def eat_pizza():
     return comment
 
 
+class Clothes:
+    pass
+
+
+def machine_wash(
+    clothes: u(
+        Clothes,
+        uri=EX.Garment,
+        triples=(EX.hasProperty, EX.MachineWashable),
+    ),
+) -> u(
+    Clothes,
+    uri=EX.SomethingElse,
+    triples=(EX.hasProperty, EX.Cleaned),
+    derived_from="inputs.clothes",
+):
+    return clothes
+
+
+@workflow
+def clothes_wf(
+    clothes: u(
+        Clothes,
+        uri=EX.Garment,
+        triples=(EX.hasProperty, EX.MachineWashable),
+    ),
+):
+    washed = machine_wash(clothes)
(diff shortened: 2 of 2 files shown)


liamhuber:
The codacy failure is the result of taking strings for `derived_from`, and not something stemming from this PR:

Screenshot 2025-08-22 at 3 06 07 PM

PR #245, opened 2025-08-25
No topo inheritance

### LINKED ISSUE #244: Don't inherit topoligical properties
In #238 I introduced a bug where connections lead ports to inherit topological information from their connection's upstream partner. This actually (rightly!) caused a test to fail, but I thoughtlessly just updated the reference string: https://github.com/pyiron/semantikon/pull/238/commits/a1cbb6f605aa5cc8b6404ea56088edcbef1c5429

As result, we get behaviour like this:

Given

```python
import owlrl
import rdflib

def upstream(x):
    return x

def downstream(y):
    return y

def simplest_wf(z):
    x = upstream(z)
    y = downstream(x)
    return y

wf_dict = workflow.get_workflow_dict(simplest_wf)
owlrl.DeductiveClosure(owlrl.OWLRL_Semantics).expand(graph)
```

We very nicely see

```python
[t for t in graph.objects(rdflib.term.URIRef("simplest_wf.downstream_0.inputs.y"), onto.SNS.inputOf)]
>>> [rdflib.term.URIRef('simplest_wf.downstream_0')]
```

But undesirably see

```python
[t for t in graph.objects(rdflib.term.URIRef("simplest_wf.downstream_0.inputs.y"), onto.SNS.outputOf)]
>>> [rdflib.term.URIRef('simplest_wf.upstream_0')]
```

Of course `"simplest_wf.downstream_0.inputs.y"` shouldn't be the `outputOf` anything!

**Proposal:** Don't inherit topological properties as part of `inheritsFrom`. These are sort of like ontological metadata and don't need to be passed along as we analyze our graph.

liamhuber:
Don't inherit topological properties `inputOf` and `outputOf` when you `inheritsProperties`.

@samwaseda, this makes a lot of sense to me -- it feels rational that an input of a node should only ever be an input of that node, never an output of anyone, and not an input of anyone else, regardless of who it's connected to. However, I created the bug in the first place, so I may be missing something critical.

Closes #244

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -248,6 +248,8 @@ def _inherit_properties(
         FILTER(?p != rdfs:label)
         FILTER(?p != rdf:value)
         FILTER(?p != ns:hasValue)
+        FILTER(?p != ns:inputOf)
+        FILTER(?p != ns:outputOf)
         FILTER(?p != owl:sameAs)
     }}
     """
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -15,7 +15,6 @@ class SNS:
     hasNode: URIRef = BASE["hasNode"]
     hasSourceFunction: URIRef = BASE["hasSourceFunction"]
     hasUnits: URIRef = BASE["hasUnits"]
-    inheritsPropertiesFrom: URIRef = BASE["inheritsPropertiesFrom"]
     inputOf: URIRef = BASE["inputOf"]
     outputOf: URIRef = BASE["outputOf"]
     hasValue: URIRef = BASE["hasValue"]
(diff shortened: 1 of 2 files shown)


samwaseda (approved):

samwaseda:
Ah wait, the intention behind this was to make sure that the output of a child node is at the same time the output of the macro node. I guess this one brings me back to my [previous comment](https://github.com/pyiron/semantikon/pull/247#issuecomment-3222971518), because the current functionality should be maintained for the direct output - input connections, but not if the user applies it via `derived_from=`

liamhuber:
> Ah wait, the intention behind this was to make sure that the output of a child node is at the same time the output of the macro node. I guess this one brings me back to my [previous comment](https://github.com/pyiron/semantikon/pull/247#issuecomment-3222971518), because the current functionality should be maintained for the direct output - input connections, but not if the user applies it via `derived_from=`

Yeah, I saw that macro/child relationship and noticed it breaking. I wonder though -- is the output of the macro the output of the child? So for instance, if I have a peer of the macro, it is certainly allowed to form connections with the output of the macro, but in principle it should be disallowed from punching through the macro walls and forming a connection with one of the macro's children. To that end, do we anywhere want to indicate that "macro output _is_ child output"? It seems to me like this may complicate enforcement of that connection rule.

liamhuber:
Ah, I saw your comment but missed your approval, @samwaseda. My response comment stands and we can keep working through all this, but I like the approval and I'll keep merging stuff down as long as I've got that and passing tests 👍

PR #247, opened 2025-08-25
Accommodate restrictions

liamhuber:
When checking validity, in `_check_connections`, we see if the downstream input has any `rdflib.RDF.type` fields that are not included in the upstream types. Here, I exclude all downstream types inheriting from `OWL.Restriction`, since we instead handle such restrictions in `_check_missing_triples`.

For the clothing example, this solves my problems and

```python
# Chain together restrictions and inheritance and restrictions
import rdflib

class Clothes:
    pass

def wash(clothes: Clothes) -> u(
    Clothes,
    uri=EX.Clothes,
    triples=(EX.hasProperty, EX.cleaned),
    derived_from="inputs.clothes"
):
    ...
    return clothes

def dye(clothes: Clothes, color="blue") -> u(
    Clothes,
    uri=EX.Clothes,
    triples=(EX.hasProperty, EX.color),
    derived_from="inputs.clothes",
):
    ...
    return clothes

def sell(
    clothes: u(
        Clothes,
        uri=EX.Clothes,
        restrictions=(
            ((rdflib.OWL.onProperty, EX.hasProperty), (rdflib.OWL.someValuesFrom, EX.cleaned)),
            ((rdflib.OWL.onProperty, EX.hasProperty), (rdflib.OWL.someValuesFrom, EX.color))
        )
    )
) -> u(Clothes, derived_from="inputs.clothes"):  # int
    ...
    return clothes

def my_correct_workflow(clothes: Clothes) -> Clothes:
    dyed_clothes = dye(clothes)
    washed_clothes = wash(dyed_clothes)
    money = sell(washed_clothes)
    more_money = sell(money)
    return more_money

wf_dict = workflow.get_workflow_dict(my_correct_workflow)
validity = onto.validate_values(graph)
validity
```

Validates fine, while if we comment out, say `triples=(EX.hasProperty, EX.cleaned),`, then the validity (correctly) fails due to a bunch of missing triples.

Thus, with #241 and this together, output that `derives_from=` the input (correctly!) captures that inputs `uri=` (and other types), _and_ we can combine `uri=` with `restrictions=`.

However, I'm still not super happy with this. First, this attack is still glaringly missing `SH`-based (instead of `OWL`-based) restrictions. But even more fundamentally, I find it quite uncomfortable that `triples=(EX.hasProperty, EX.cleaned)` doesn't impact the `RDF.type` set at all, and yet it satisfies `restrictions=((rdflib.OWL.onProperty, EX.hasProperty), (rdflib.OWL.someValuesFrom, EX.cleaned))`, which has all these extra terms and _does_ modify the set of types! So this gets the behaviour I want, but in a way that makes me uncomfortable. With that said, this is a complicated thing and I am not sure I have a better answer right now.

@samwaseda could you take a critical look at this attack and tell me what you think?

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -294,16 +294,16 @@ def _check_connections(graph: Graph, strict_typing: bool = False) -> list:
         (list): list of incompatible connections
     """
     incompatible_types = []
-    for inp_out in graph.subject_objects(SNS.inheritsPropertiesFrom):
+    for (inp, out) in graph.subject_objects(SNS.inheritsPropertiesFrom):
         i_type, o_type = [
             [g for g in graph.objects(tag, RDF.type) if g != PROV.Entity]
-            for tag in inp_out
+            for tag in (inp, out)
         ]
         if not strict_typing and (i_type == [] or o_type == []):
             continue
         diff = set(i_type).difference(o_type)
         if len(diff) > 0:
-            incompatible_types.append(inp_out + (i_type, o_type))
+            incompatible_types.append((inp, out) + (i_type, o_type))
     return incompatible_types
 
 
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -279,7 +279,7 @@ def _check_missing_triples(graph: Graph) -> list:
             ?instance ?onProperty ?someValuesFrom .
         }
     }"""
-    return list(graph.query(query))
+    return list(set(graph.query(query)))
 
 
 def _check_connections(graph: Graph, strict_typing: bool = False) -> list:
@@ -299,9 +299,20 @@ def _check_connections(graph: Graph, strict_typing: bool = False) -> list:
             [g for g in graph.objects(tag, RDF.type) if g != PROV.Entity]
             for tag in (inp, out)
         ]
-        if not strict_typing and (i_type == [] or o_type == []):
+        # Exclude any i_type that is an OWL restriction or a subclass of OWL.Restriction
+        # (What about SH restrictions?)
+        # This is because we handle restrictions in _check_missing_triples
+        i_type_filtered = [
+            t
+            for t in i_type
+            if (
+                (t, RDFS.subClassOf, OWL.Restriction) not in graph
+                and (t, RDF.type, OWL.Restriction) not in graph
+            )
+        ]
+        if not strict_typing and (i_type_filtered == [] or o_type == []):
             continue
-        diff = set(i_type).difference(o_type)
+        diff = set(i_type_filtered).difference(o_type)
         if len(diff) > 0:
             incompatible_types.append((inp, out) + (i_type, o_type))
     return incompatible_types
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -308,6 +308,8 @@ def _check_connections(graph: Graph, strict_typing: bool = False) -> list:
             if (
                 (t, RDFS.subClassOf, OWL.Restriction) not in graph
                 and (t, RDF.type, OWL.Restriction) not in graph
+                and (t, RDFS.subClassOf, SH.NodeShape) not in graph
+                and (t, RDF.type, SH.NodeShape) not in graph
             )
         ]
         if not strict_typing and (i_type_filtered == [] or o_type == []):
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -248,6 +248,8 @@ def _inherit_properties(
         FILTER(?p != rdfs:label)
         FILTER(?p != rdf:value)
         FILTER(?p != ns:hasValue)
+        FILTER(?p != ns:inputOf)
+        FILTER(?p != ns:outputOf)
         FILTER(?p != owl:sameAs)
     }}
     """
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -232,30 +232,30 @@ def _parse_triple(
 def _inherit_properties(
     graph: Graph, triples_to_cancel: list | None = None, n_max: int = 1000, ontology=SNS
 ):
-    update_query = (
-        f"PREFIX ns: <{ontology.BASE}>",
-        f"PREFIX rdfs: <{RDFS}>",
-        f"PREFIX rdf: <{RDF}>",
-        f"PREFIX owl: <{OWL}>",
-        "",
-        "INSERT {",
-        "    ?subject ?p ?o .",
-        "}",
-        "WHERE {",
-        "    ?subject ns:inheritsPropertiesFrom ?target .",
-        "    ?target ?p ?o .",
-        "    FILTER(?p != ns:inheritsPropertiesFrom)",
-        "    FILTER(?p != rdfs:label)",
-        "    FILTER(?p != rdf:value)",
-        "    FILTER(?p != ns:hasValue)",
-        "    FILTER(?p != owl:sameAs)",
-        "}",
-    )
+    update_query = f"""\
+    PREFIX ns: <{ontology.BASE}>
+    PREFIX rdfs: <{RDFS}>
+    PREFIX rdf: <{RDF}>
+    PREFIX owl: <{OWL}>
+    
+    INSERT {{
+        ?subject ?p ?o .
+    }}
+    WHERE {{
+        ?subject ns:inheritsPropertiesFrom ?target .
+        ?target ?p ?o .
+        FILTER(?p != ns:inheritsPropertiesFrom)
+        FILTER(?p != rdfs:label)
+        FILTER(?p != rdf:value)
+        FILTER(?p != ns:hasValue)
+        FILTER(?p != owl:sameAs)
+    }}
+    """
     if triples_to_cancel is None:
         triples_to_cancel = []
     n = 0
     for _ in range(n_max):
-        graph.update("\n".join(update_query))
+        graph.update(update_query)
         for t in triples_to_cancel:
             if t in graph:
                 graph.remove(t)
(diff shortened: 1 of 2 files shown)


semantikon/ontology.py
```diff
@@ -294,16 +294,27 @@ def _check_connections(graph: Graph, strict_typing: bool = False) -> list:
...
-        if not strict_typing and (i_type == [] or o_type == []):
+        # Exclude any i_type that is an OWL restriction or a subclass of OWL.Restriction
+        # (What about SH restrictions?)
+        # This is because we handle restrictions in _check_missing_triples
+        i_type_filtered = [
+            t
+            for t in i_type
+            if (
+                (t, RDFS.subClassOf, OWL.Restriction) not in graph
+                and (t, RDF.type, OWL.Restriction) not in graph
```
samwaseda, line 310:
```suggestion
                (t, RDFS.subClassOf, OWL.Restriction) not in graph
                and (t, RDF.type, OWL.Restriction) not in graph
                and (t, RDFS.subClassOf, SH.NodeShape) not in graph
                and (t, RDF.type, SH.NodeShape) not in graph
```

liamhuber:
> ...Maybe we should introduce another predicate for this?

This whole comment is very useful to me. And I think it is important to the issue of how we're going to handle restrictions. When I first started writing, I intended to suggest that we need no further predicate... but I managed to convince myself otherwise 😂 

For inter-node edges, we are basically trying to guarantee that the upstream entity (the `output_A`) is ontologically "as or more specific" than the downstream receiver (`input_B`). In contrast, for intra-node edges we want the downstream (`output_B`) to definitively inherit everything from the upstream (`input_B`), and it can't have anything explicitly in conflict with the upstream, but it _can_ add more detail. I.e. we might get in a `EX.Dinosaur`, and get out a `EX.Triceratops` or something.

With this in mind, I set out to try and break some things. It was interesting to me that when we extend output by demanding new non-type `triples=` in the output, this consistently works fine (on this branch! on `main` we'd get restrictions causing type disjointness errors). This is because our "as or more specific than" check is only looking at (non-restriction) `rdflib.RDF.type` relationships, which doesn't look at the new property triples we're slapping on. The restrictions infrastructure makes sure these get converted into types later! So the current restrictions setup is letting us _bypass_ the regular inheritance check in the intra-node edge case! But this all depends on the relationship between how `triples=` and `restrictions=` get parsed, which feels very fragile to me.

<details>
  <summary>Example with adding properties</summary>
  
```python
# A toy example for inter- vs intra-node edges
import rdflib

class Person: ...

def mutate(hero: u(Person, uri=EX.Hero)) -> u(
    Person,
    derived_from="inputs.hero",
    triples=(EX.hasPower, EX.SuperStrength),
):
    return hero

def train(trainee: u(
        Person,
        uri=EX.Hero,
        restrictions=(
            (
                (rdflib.OWL.onProperty, EX.hasPower),
                (rdflib.OWL.someValuesFrom, EX.SuperStrength)
            ),
        ),
    )
) -> u(
    Person,
    derived_from="inputs.trainee",
    triples=(EX.hasPower, EX.SuperSpeed),
):
    return trainee

def battle(challenger: u(
        Person,
        uri=EX.Hero,
        restrictions=(
            (
                (rdflib.OWL.onProperty, EX.hasPower),
                (rdflib.OWL.someValuesFrom, EX.SuperStrength)
            ),
            (
                (rdflib.OWL.onProperty, EX.hasPower),
                (rdflib.OWL.someValuesFrom, EX.SuperSpeed)
            )
        )
    )
):
    return "Victory!"

def video_game(character):
    protagonist = mutate(character)
    hero = train(protagonist)
    outcome = battle(hero)
    return outcome

wf_dict = workflow.get_workflow_dict(video_game)
validity = onto.validate_values(graph)
validity
>>> {'missing_triples': [], 'incompatible_connections': []}
```

I.e. we're able to add triples to output that derives from the input.

```python
[
    x for x in graph.objects(
        rdflib.term.URIRef('video_game.battle_0.inputs.challenger'),
        rdflib.RDF.type,
    )
]
>>> [rdflib.term.URIRef('http://www.w3.org/ns/prov#Entity'),
... rdflib.term.URIRef('http://www.example.org/Hero'),
... rdflib.term.URIRef('Ndc2428ad2b8e49a0b6b0a1c182c46d79'),
... rdflib.term.URIRef('N22d1c39ef3ce48bb82ff95d38e67f565'),
... rdflib.term.URIRef('N99a5af785f86433ea2fb3cfacd68cf27')]
```

One tangential issue I have is that this should really have _two_ restrictions (strength and speed), but somehow we wind up with three. I guess the conversion code from `restrictions=` to actual types is not correctly identifying that the restriction is already in place?
  
</details>

Ok, but I really _wanted_ to break something. One way to accomplish this is by brute-force violating our own typing laws:

```python
import rdflib

def geppetto(puppet: u(str, uri=EX.Puppet)) -> u(
    str,
    derived_from="inputs.puppet",
    triples=(rdflib.RDF.type, EX.Sentient),
):
    return puppet

def story(name: str = "pinocchio"):
    protagonist = geppetto(name)
    return protagonist

wf_dict = workflow.get_workflow_dict(story)
validity = onto.validate_values(graph)
validity
>>> {'missing_triples': [],
... 'incompatible_connections': [(rdflib.term.URIRef('story.geppetto_0.outputs.puppet'),
...    rdflib.term.URIRef('story.geppetto_0.inputs.puppet'),
...    [rdflib.term.URIRef('http://www.example.org/Sentient'),
...     rdflib.term.URIRef('http://www.example.org/Puppet')],
...    [rdflib.term.URIRef('http://www.example.org/Puppet')])]}
```

So maybe we need to re-think the "all classes are inherently disjoint" argument, because this seems like a perfectly reasonable step.

<details>
  <summary>We can also frame it slightly differently to the same failing result</summary>

```python
import rdflib

def geppetto(puppet: u(str, uri=EX.Puppet)) -> u(
    str,
    uri=EX.MagicPuppet,
    derived_from="inputs.puppet",
):
    return puppet

def story(name: str = "pinocchio"):
    protagonist = geppetto(name)
    return protagonist

g0 = rdflib.Graph()
g0.add((EX.MagicPuppet, rdflib.RDFS.subClassOf, EX.Puppet))
g0.add((EX.MagicPuppet, rdflib.RDFS.subClassOf, EX.Sentient))

wf_dict = workflow.get_workflow_dict(story)
graph = onto.get_knowledge_graph(wf_dict, graph=g0)
validity = onto.validate_values(graph)
validity
>>> {'missing_triples': [],
...  'incompatible_connections': [(rdflib.term.URIRef('story.geppetto_0.outputs.puppet'),
...    rdflib.term.URIRef('story.geppetto_0.inputs.puppet'),
...    [rdflib.term.URIRef('http://www.example.org/MagicPuppet'),
...     rdflib.term.URIRef('http://www.example.org/Puppet'),
...     rdflib.term.URIRef('http://www.example.org/Sentient')],
...    [rdflib.term.URIRef('http://www.example.org/Puppet')])]}
```
  
</details>

So I wonder if we need to rethink the inheritance and connection validation?

liamhuber:
> Not sure if that's true, but maybe we can look for them

We might be able to add and exploit this, but if by "look for" you meant see if they're already there and exploit them, they're not. I took one of the clothing examples and ran

```python
(
    [x for x in graph.subject_objects(rdflib.RDFS.Class)],
    [x for x in graph.subject_objects(rdflib.OWL.Class)],
    [x for x in graph.subject_predicates(rdflib.RDFS.Class)],
    [x for x in graph.subject_predicates(rdflib.OWL.Class)],
)
```

Only the final search `(s, rdflib.OWL.Class, o)` had non-zero results and they are not interesting:

```
 [(rdflib.term.URIRef('http://www.w3.org/2002/07/owl#Thing'),
   rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type')),
  (rdflib.term.URIRef('http://www.w3.org/2002/07/owl#Nothing'),
   rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type')),
  (rdflib.term.URIRef('http://www.w3.org/2002/07/owl#Class'),
   rdflib.term.URIRef('http://www.w3.org/2002/07/owl#sameAs'))])
```

On the other hand, if you mean that we can insist that for `uri=`, `triples=`, `restrictions=`, we leverage one of these classes (and insist that users call them instead of `rdflib.RDF.type` to indicate type-ness), while leaving plain `rdflib.RDF.type` for the topological types (#245) and anything else -- then indeed maybe it can be a useful tool.

liamhuber:
Dammit, I thought I was on a different branch. Sloppy push.

liamhuber:
> Dammit, I thought I was on a different branch. Sloppy push.

I reset HEAD and force-pushed to clean up.

liamhuber:
@samwaseda, IMO this is superseded by #252. I branched that off here, so I want to leave this around so you can look at the diff more cleanly to see what was needed to get `SNS.fulfills` working, but if you like the direction then I would point #252 directly at main (including the commits from this branch) and close this PR and delete this branch.

samwaseda (commented):
I have actually a slightly different suggestion: Currently we are comparing the types belonging to the same node. But how about comparing the type of the output argument and that of the connected input argument? At the end of the day, it requires only some changes in the SPARQL query.

samwaseda:
Sorry, I forgot that we didn't have the SPARQL query anymore, and yes, in that case we would use a SPARQL query instead of `owlrl`. Now I cannot remember anymore why I made this suggestion... Let me see if I remember it before you wake up :D

samwaseda:
> In general, I actually dislike the `"missing_triples"` field relative to the `"incompatible_connections"` field, as the latter lets me isolate exactly where things went wrong, where the former just tells me one of the suspects involved in the crime.

That's a very valid point. Let me work on it

samwaseda:
ok let's merge it now XD

PR #252, opened 2025-08-26
`fulfills` predicate as a counterpart to `wasDerivedFrom`

liamhuber:
Thinking through a bunch of the challenges in #247, with particular weight on the ideas in [@samwaseda's comment](https://github.com/pyiron/semantikon/pull/247#issuecomment-3222971518) about inter- vs intra-node ontological connections and the idea of an additional predicate, I try here to introduce a new `sns:fulfills` predicate.

This rests on two key concepts. The first is the hypothesis that we can split apart "true"[^1] and "bookkeeping" ontological information. I'm not confident enough in my ontology experience to be confident that this is without failing edge cases, but it's drawn from `ontology._inherit_properties`, which has been pretty successful so far. The idea there is to filter out all the junk we've added in, and leave only ontological information that the user is actually trying to add to their workflows. (Cf. also #244.) 
  I think this is worth highlighting since this is a potential failure point, but it's really super simple: we're just writing an explicit list of terms to exclude when comparing the sets of triples defining two entities.

[^1]: Or "interesting", whatever. The point is they're the ones we care about that the user has indicated matter.

The second is that we can break apart the concept of `prov:wasDerivedFrom` ensuring that ontological information gets carried from upstream to downstream, and the concept that the connections carrying that information may or may not be valid depending on the annotations of the partners in the connection, their context, and the direction the flow of information. [@samwaseda's comment](https://github.com/pyiron/semantikon/pull/247#issuecomment-3222971518) is the canonical example here. 
  What we need is to re-pose the `pyiron_workflow` concept of [type_hint_is_as_or_more_specific_than](https://github.com/pyiron/pyiron_workflow/blob/dc22e3555d5029ee7fc545fb70d43a3aca5c9bd0/pyiron_workflow/type_hinting.py#L62), but for ontologies. For this, I introduced the new predicate `SNS.fulfills`. 
  For inter-node connections, fulfillment and inheritance run opposite[^2] each other: `(upstream, SNS.fulfills, downstream)` and `(downstream, PROV.wasDerivedFrom, upstream)`. I.e., we should guarantee that downstream carries all the information from upstream (derivation) and that downstream doesn't require anything missing from that upstream inheritance (i.e. fulfillment, or upstream is as-or-more-specific-than downstream). 
  For intra-node connections, fulfillment and inheritance run together:  `(downstream, SNS.fulfills, upstream)` and `(downstream, PROV.wasDerivedFrom, upstream)`. I.e., the output of a node that derives from some input field should (unless a cancellation is explicitly noted) carry all the information of the upstream partner _and_ it is free to add additional information.

[^2]: Or this is running in the same direction and intra-nodal connections run opposite. I don't care, the point is just that there is a polarity difference between inter- and intra-nodal connections.

Putting these two together, if `(a, PROV.wasDerivedFrom, b)`, then we want to copy over all the "true" triples from `b` to `a`; if `(c, SNS.fulfills, d)`, then the "true" triples of `d` are a strict subset of the "true" triples of `c`. In this way we propagate information and can validate types without (necessarily) worrying about `restrictions=` or the extra restrictions classes (although we still need `cancel=`).

As of 9d5c1fc060bddd7dd4fea6357c6b9eaf676e0165, I've implemented this fully in parallel to the restrictions approach. That means all the tests are still running (with the changes in #247), and the only difference is that `ontology.validate_values` includes an extra field for parsing if we've broken any `SNS.fulfills` promises.

Critical to check here, but I haven't gotten to:
- Work with `cancel` to allow intra-node connections to _remove_ information vis-a-vis what they `PROV.wasDerivedFrom`
- Check that explicit user restrictions, like `OWL.disjointWith` or a `OWL.Restriction` still result in a validation failure

With that, I wanted to revisit my example from [this comment](https://github.com/pyiron/semantikon/pull/247#issuecomment-3224511243) to demonstrate how the new attack presently[^3] looks.

[^3]: 9d5c1fc060bddd7dd4fea6357c6b9eaf676e0165

## The original way with restrictions

With restrictions, we would write

```python
import rdflib

class Person: ...

def mutate(hero: u(Person, uri=EX.Hero)) -> u(
    Person,
    derived_from="inputs.hero",
    triples=(EX.hasPower, EX.SuperStrength),
):
    return hero

def train(trainee: u(
        Person,
        uri=EX.Hero,
        restrictions=(
            (
                (rdflib.OWL.onProperty, EX.hasPower),
                (rdflib.OWL.someValuesFrom, EX.SuperStrength)
            ),
        ),
    )
) -> u(
    Person,
    derived_from="inputs.trainee",
    triples=(EX.hasPower, EX.SuperSpeed),
):
    return trainee

def battle(challenger: u(
        Person,
        uri=EX.Hero,
        restrictions=(
            (
                (rdflib.OWL.onProperty, EX.hasPower),
                (rdflib.OWL.someValuesFrom, EX.SuperStrength)
            ),
            (
                (rdflib.OWL.onProperty, EX.hasPower),
                (rdflib.OWL.someValuesFrom, EX.SuperSpeed)
            )
        )
    )
):
    return "Victory!"

def video_game(character):
    protagonist = mutate(character)
    hero = train(protagonist)
    outcome = battle(hero)
    return outcome

wf_dict = workflow.get_workflow_dict(video_game)
validity = onto.validate_values(graph)
validity
>>> {'missing_triples': [],
...  'incompatible_connections': [],
...  'broken_promises': {rdflib.term.URIRef('video_game.mutate_0.outputs.hero'): 
... (rdflib.term.URIRef('video_game.train_0.inputs.trainee'),
...    {(rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'),
...      rdflib.term.URIRef('N2adc791ac25c4950ae927729262e0200')),
...     (rdflib.term.URIRef('http://www.w3.org/1999/02/22-rdf-syntax-ns#type'),
...      rdflib.term.URIRef('N886e6081649e419cb843187e3a6cd8d3'))}),
...   rdflib.term.URIRef('video_game.train_0.outputs.trainee'): ...
...  ....
```

We've been correctly using restrictions, so there are no "incompatible connections", but the current restricitons setup runs around introducing new `RDF.type` to our entities, so the fulfillment validation complains left, right, and center. That's not so important, although I do want to take a moment to highlight the dictionary `fulfiller: (fulfilled, (unfulfilled predictate-object))` format for the problems, which I think has improved readability compared to "incompatible_connections" list approach.

## The new way for fulfillment promises

```python
import rdflib

class Person: ...

def mutate(hero: u(Person, uri=EX.Hero)) -> u(
    Person,
    derived_from="inputs.hero",
    triples=(EX.hasPower, EX.SuperStrength),
):
    return hero

def train(trainee: u(
        Person,
        uri=EX.Hero,
        triples=(EX.hasPower, EX.SuperStrength),
    )
) -> u(
    Person,
    derived_from="inputs.trainee",
    triples=(EX.hasPower, EX.SuperSpeed),
):
    return trainee

def battle(challenger: u(
        Person,
        uri=EX.Hero,
        triples=(
            (EX.hasPower, EX.SuperStrength),
            (EX.hasPower, EX.SuperSpeed),
        )
    )
):
    return "Victory!"

def video_game(character):
    protagonist = mutate(character)
    hero = train(protagonist)
    outcome = battle(hero)
    return outcome

wf_dict = workflow.get_workflow_dict(video_game)
validity = onto.validate_values(graph)
validity
>>> {'missing_triples': [], 'incompatible_connections': [], 'broken_promises': {}}
```

IMO this is easier to work with than the restrictions. I always found it very confusing that if my upstream `(EX.hasPower, EX.SuperStrength)`, then my downstream needed to guarantee that it had a property `EX.hasPower` with "some values from" `EX.SuperStrength`. Now, if my input wants the triple `(EX.hasPower, EX.SuperStrength)`, then my upstream had better have that!

Note how `train` demands `(EX.hasPower, EX.SuperStrength)` for its input, inherits from that input, adds the triple `(EX.hasPower, EX.SuperSpeed)`, and consequently has both at output.

Suppose now that we go in and modify `train` so it doesn't add this extra triple:

```python
def train(trainee: u(
        Person,
        uri=EX.Hero,
        triples=(EX.hasPower, EX.SuperStrength),
    )
) -> u(
    Person,
    derived_from="inputs.trainee",
    # triples=(EX.hasPower, EX.SuperSpeed),
):
    return trainee
```

In this case our validation comes out as 

```python
{
    "missing_triples": [],
    "incompatible_connections": [],
    "broken_promises": {
        rdflib.term.URIRef("video_game.train_0.outputs.trainee"): (
            rdflib.term.URIRef("video_game.battle_0.inputs.challenger"),
            {
                (
                    rdflib.term.URIRef("http://www.example.org/hasPower"),
                    rdflib.term.URIRef("http://www.example.org/SuperSpeed"),
                )
            },
        )
    },
}
```

I'm not worried that "incompatible_connections" is empty -- that needs restrictions and we haven't provided those --  but we see exactly what we want for the fulfillment: the output of `train` is missing `SuperSpeed` for the input of `battle`!

I'll push more examples and edge cases to see if this has any nasty failure points, but so far I'm pretty happy and optimistic. I'm looking forward to your feedback, @samwaseda

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -2,6 +2,7 @@ import warnings
 from dataclasses import is_dataclass
 from typing import Any, Callable, TypeAlias, cast
 
+import rdflib.term
 from owlrl import DeductiveClosure, OWLRL_Semantics
 from rdflib import OWL, PROV, RDF, RDFS, SH, BNode, Graph, Literal, Namespace, URIRef
 from rdflib.term import IdentifiedNode
@@ -110,7 +111,7 @@ def _get_triples_from_restrictions(data: dict) -> list:
         triples.extend(
             [
                 ("self", PROV.wasDerivedFrom, data["derived_from"]),
-                ("self", SNS.fulfills, data["derived_from"])
+                ("self", SNS.fulfills, data["derived_from"]),
             ]
         )
     return triples
@@ -327,6 +328,49 @@ def _check_connections(graph: Graph, strict_typing: bool = False) -> list:
     return incompatible_types
 
 
+def _query_meaning(
+    graph: Graph, subject: rdflib.term.Node, ontology=SNS
+):
+    query = f"""\
+        PREFIX ns: <{ontology.BASE}>
+        PREFIX prov: <{PROV}>
+        PREFIX rdfs: <{RDFS}>
+        PREFIX rdf: <{RDF}>
+        PREFIX owl: <{OWL}>
+        
+        SELECT ?p ?o 
+        
+        WHERE {{
+            ?s ?p ?o .
+            FILTER(?s = <{subject}>)
+            FILTER(?p != prov:wasDerivedFrom)
+            FILTER(?p != rdfs:label)
+            FILTER(?p != rdf:value)
+            FILTER(?p != ns:hasValue)
+            FILTER(?p != ns:inputOf)
+            FILTER(?p != ns:outputOf)
+            FILTER(?p != ns:fulfills)
+            FILTER(?p != owl:sameAs)
+            FILTER(!(?p = rdf:type && ?o = prov:Entity))  # Null info added ubiquitously
+        }}
+    """
+    return set(graph.query(query))
+
+
+def _check_fulfillment(graph: Graph, strict_typing: bool = False):
+    promises = graph.subject_objects(SNS.fulfills)
+    broken_promises = {}
+    for fulfilled_by, fulfilled in promises:
+        alleged_superset = _query_meaning(graph, fulfilled_by)
+        alleged_subset = _query_meaning(graph, fulfilled)
+        if not strict_typing and (not alleged_superset or not alleged_subset):
+            continue
+        diff = alleged_subset.difference(alleged_superset)
+        if len(diff) > 0:
+            broken_promises[fulfilled_by] = fulfilled, diff
+    return broken_promises
+
+
 def validate_values(
     graph: Graph, run_reasoner: bool = True, strict_typing: bool = False
 ) -> dict[str, list]:
@@ -348,6 +392,9 @@ def validate_values(
         "incompatible_connections": _check_connections(
             graph, strict_typing=strict_typing
         ),
+        "broken_promises": _check_fulfillment(
+            graph, strict_typing=strict_typing
+        )
     }
 
 
@@ -466,7 +513,7 @@ def _edges_to_triples(edges: dict) -> list:
         triples.extend(
             [
                 (downstream, PROV.wasDerivedFrom, upstream),
-                (upstream, SNS.fulfills, downstream)
+                (upstream, SNS.fulfills, downstream),
             ]
         )
     return triples
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -328,9 +328,7 @@ def _check_connections(graph: Graph, strict_typing: bool = False) -> list:
     return incompatible_types
 
 
-def _query_meaning(
-    graph: Graph, subject: rdflib.term.Node, ontology=SNS
-):
+def _query_meaning(graph: Graph, subject: rdflib.term.Node, ontology=SNS):
     query = f"""\
         PREFIX ns: <{ontology.BASE}>
         PREFIX prov: <{PROV}>
@@ -392,9 +390,7 @@ def validate_values(
         "incompatible_connections": _check_connections(
             graph, strict_typing=strict_typing
         ),
-        "broken_promises": _check_fulfillment(
-            graph, strict_typing=strict_typing
-        )
+        "broken_promises": _check_fulfillment(graph, strict_typing=strict_typing),
     }
 
 
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -18,6 +18,7 @@ class SNS:
     inputOf: URIRef = BASE["inputOf"]
     outputOf: URIRef = BASE["outputOf"]
     hasValue: URIRef = BASE["hasValue"]
+    fulfills: URIRef = BASE["fulfills"]
 
 
 class NS:
@@ -250,6 +251,7 @@ def _inherit_properties(
         FILTER(?p != ns:hasValue)
         FILTER(?p != ns:inputOf)
         FILTER(?p != ns:outputOf)
+        FILTER(?p != ns:fulfills)
         FILTER(?p != owl:sameAs)
     }}
     """
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -107,7 +107,12 @@ def _get_triples_from_restrictions(data: dict) -> list:
     if data.get("triples", None) is not None:
         triples.extend(_align_triples(data["triples"]))
     if data.get("derived_from", None) is not None:
-        triples.append(("self", PROV.wasDerivedFrom, data["derived_from"]))
+        triples.extend(
+            [
+                ("self", PROV.wasDerivedFrom, data["derived_from"]),
+                ("self", SNS.fulfills, data["derived_from"])
+            ]
+        )
     return triples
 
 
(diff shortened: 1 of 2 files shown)


liamhuber:
> Check that explicit user restrictions, like `OWL.disjointWith` or a `OWL.Restriction` still result in a validation failure

I think there is no way to cause these explicit restrictions to fail unless we explicitly check for them. I think a good attack for that will be extending `ontology.validate_values` to take arguments for post-expansion graph validation routines. We may want to include some standard subset of these by default, e.g. I chatted with Claude to get a SPARQL query for managing `OWL.disjointWith` checks the raw graph level

```python
import owlrl
import rdflib

g = rdflib.Graph()
g.add((EX.Foo, rdflib.RDFS.subClassOf, EX.Bar))
g.add((EX.Bar, rdflib.RDFS.subClassOf, EX.NotFoo))
g.add((EX.Foo, rdflib.OWL.disjointWith, EX.NotFoo))
owlrl.DeductiveClosure(owlrl.OWLRL_Semantics).expand(g)

simple_disjoint_violations_query = f"""\
PREFIX owl: <{rdflib.OWL}>
PREFIX rdfs: <{rdflib.RDFS}>

SELECT ?issue_type ?class1 ?class2 WHERE {{
    ?class1 rdfs:subClassOf ?class2 .
    ?class1 owl:disjointWith ?class2 .
    BIND("disjoint" AS ?issue_type)
}}
"""
violations = list(g.query(simple_disjoint_violations_query))
violations
>>> [(rdflib.term.Literal('disjoint'),
...   rdflib.term.URIRef('http://www.example.org/Foo'),
...   rdflib.term.URIRef('http://www.example.org/NotFoo'))]
```

Doing something like an optional dictionary with string keys and items like (callable, args, kwargs) would give a nice way to include the results in the validation report (under the same string)

liamhuber:
Ok, I'm happy with this as a new alternative to `restrictions=`. I've duplicated all the existing tests for restrictions, so I think this is ready to go.

On my TODO-list for tomorrow (if @samwaseda likes this PR) but in stacked/subsequent PRs not directly here:
- Update `knowledge_graph.ipynb` to use the direct `triples=` approach here instead of `restrictions=`
- Remove `restrictions=` from the `u` signature[^1]
- Remove the `restrictions` code altogether
- Add `cancel=` to the `u` signature[^1]
- Add a default graph validation for `OWL.disjointWith`[^2]
  - @samwaseda, any other patterns we should scan by default?
- Refactor `validate_values` to optionally take additional graph validator routines from the user 

[^1]: https://github.com/pyiron/semantikon/pull/252#issuecomment-3225941751
[^2]: https://github.com/pyiron/semantikon/pull/252#issuecomment-3225911587

liamhuber:
Ought "fulfills" mean that in addition to the subject, the fulfiller also satisfies all relationships in which the fulfilled is the object?

My knee jerk reaction is "probably yes", but I can't think of a concrete example off hand.

samwaseda:
I have a bit of concerns:

### `sns:fulfills`

Now with this implementation the difference between inter-node edges and intra-node edges is understood by the difference in directions of `prov:wasDerivedFrom` and `sns:fulfills`. As I said in [this comment](https://github.com/pyiron/semantikon/pull/247#issuecomment-3222971518), however, I have the feeling that the two types of connections are actually fundamentally different, so I might actually prefer to not use `prov:wasDerivedFrom` for the inter-node edges and instead invent a new predicate for this. That means `prov:wasDerivedFrom` would be reserved only for `derived_from` as given in `u`.

### Restrictions

One thing I really like about the old implementation of `restrictions` is that it closely follows the OWL syntax. A problem I see with the new implementation is that the meaning of `triples=` becomes less clear, because you can imagine a case like:

```python
def create_atoms(element: u(str, triples=(EX.elementOf, "output.structure"))) -> Atoms:
    ...
    return structure
```

In this case `EX.elementOf` - `"output.structure"` has nothing to do with restrictions, but purely gives additional metadata to the knowledge graph.

On top of this, this PR makes the syntax for the restrictions easier, but the example was based on my mock-up code which may or may not be applicable in real science. In addition, as we all know I don't really know much about ontologies, so I might have used the restrictions wrongly, so I would actually wait for the real users to use the restrictions before going for more simplification.

PR #278, opened 2025-10-14
Use properties from PMDco and co

samwaseda:
Following discussions with @joergwa I made the first changes:

## Important Changes

- [x] Source functions are `iao:informationContentEntity`
- [x] `sns:hasSourceFunction` is replaced by `iao:isAbout` (with subject and object switched)
- [x] `sns:hasNode` is replaced by `ro:hasPart`
- [x] `sns:hasUnits` is replaced by `qudt:hasUnit`
- [x] Introduce `bfo:precedes`
- [x] Replace `sns:hasValue` by `ro:participatesIn`
- [x] Replace `sns:hasInput` and `sns:hasOutput` by `bfo:hasPart`

PMDCO

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -503,6 +503,20 @@ def _edges_to_triples(edges: dict, ontology=SNS) -> list:
     ]
 
 
+def _get_precedes(
+    edge_dict: dict[str, str], ontology=SNS
+) -> list[tuple[str, URIRef, str]]:
+    triples = []
+    for dest, prov in edge_dict.items():
+        if min(len(dest.split(".")), len(prov.split("."))) < 3:
+            continue
+        dest_node, dest_io, _ = dest.rsplit(".", 2)
+        prov_node, prov_io, _ = prov.rsplit(".", 2)
+        if dest_io == "inputs" and prov_io == "outputs":
+            triples.append((prov_node, SNS.precedes, dest_node))
+    return triples
+
+
 def _parse_workflow(
     node_dict: dict,
     channel_dict: dict,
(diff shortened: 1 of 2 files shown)


samwaseda:
@joergwa: I find it somewhat confusing that `bfo:has_part` is now used to connect the workflow and the child nodes, as well as the node and the ports, meaning:

- `my_workflow` - `has_part` - `node_1`
- `node_1` - `has_part` - `inputs.x`

Do you see a possibility to use two different properties for them?

joergwa:
> @joergwa: I find it somewhat confusing that `bfo:has_part` is now used to connect the workflow and the child nodes, as well as the node and the ports, meaning:
> 
>     * `my_workflow` - `has_part` - `node_1`
> 
>     * `node_1` - `has_part` - `inputs.x`
> 
> Do you see a possibility to use two different properties for them?

Assuming that `my_workflow` is a [process](http://purl.obolibrary.org/obo/BFO_0000015) and `node_1` also is, the first `has_part` is OK.  
Since `inputs.x` is not a process, but - I guess - some [information content entity](http://purl.obolibrary.org/obo/IAO_0000030), it cannot be part of a process.

Because (from the editorial note of [has part](http://purl.obolibrary.org/obo/BFO_0000051) in [RO](http://purl.obolibrary.org/obo/ro.owl)):

> Parthood requires the part and the whole to have compatible classes: only an occurrent have an occurrent as part; only a process can have a process as part; only a continuant can have a continuant as part; only an independent continuant can have an independent continuant as part; only a specifically dependent continuant can have a specifically dependent continuant as part; only a generically dependent continuant can have a generically dependent continuant as part. (This list is not exhaustive.)
> 
> A continuant cannot have an occurrent as part: use 'participates in'. An occurrent cannot have a continuant as part: use 'has participant'. An immaterial entity cannot have a material entity as part: use 'location of'. ...

So I suggest: 
 * `node_1` - [has participant](http://purl.obolibrary.org/obo/RO_0000057) - `inputs.x`

PR #298, opened 2025-11-23
Add identifier

samwaseda:
I forgot to add an identifier (suggested by @liamhuber on Matrix)


samwaseda:
I realized that this can be merged only after a new release of `flowrep` is made.

PR #341, opened 2026-01-03
Update numpy to 2.4.0

jan-janssen:


samwaseda (approved):

jan-janssen:
Does it make sense to release a new version, or should we wait for the 1.0 release?

samwaseda:
It's ready for 1.0 except for the documentation. If there's no urgent reason I would first finish the last bit and then release 1.0

PR #422, opened 2026-04-10
Always prepend to hash dict keys

### LINKED ISSUE #421: Node labels have secret, strictly enforced edge case
Consider the first example in the ontology notebook, but actually run it:

```python
from rdflib import Namespace

from semantikon import get_knowledge_graph, u, validate_values

EX = Namespace("http://www.example.org/")


def get_speed(
    distance: u(float, uri=EX.distance, units="meter"),
    time: u(float, uri=EX.time, units="second")
) -> u(float, uri=EX.velocity, units="meter/second"):
    speed = distance / time
    return speed


def get_kinetic_energy(
    mass: u(float, uri=EX.mass, units="kilogram"),
    velocity: u(float, uri=EX.velocity, units="meter/second")
) -> u(float, uri=EX.kineticEnergy, units="joule"):
    return 0.5 * mass * velocity**2


@workflow
def my_workflow(distance, time, mass):
    speed = get_speed(distance, time)
    kinetic_energy = get_kinetic_energy(mass, speed)
    return kinetic_energy


workflow_dict = my_workflow.run(distance=1.0, time=2.0, mass=3.0)

graph = get_knowledge_graph(workflow_dict)

validate_values(graph)

```

This proceeds perfectly fine. But, `get_knowledge_graph` is exposed in the public API, and we might reasonably have obtained our `wokflow_dict` from some other place. Suppose now that we did that, and suppose that we made the innocuous choice of having our parent node and child node have the same label -- a perfectly reasonable choice since each is still uniquely identifiable in it's scope, it is only siblings which cannot share the same label. In such a case, we wind up with an unfortunate and opaque error when trying to construct the knowledge graph:

```python
import copy

my_wf_dict = copy.deepcopy(workflow_dict)
first_child_label = next(iter(my_wf_dict["nodes"].keys()))
my_wf_dict["label"] = first_child_label
my_graph = get_knowledge_graph(my_wf_dict)
validate_values(my_graph)

```

Yields from `get_knowledge_graph(my_wf_dict)`

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File ~/dev/miniforge3/envs/pyiron12/lib/python3.12/site-packages/networkx/classes/digraph.py:926, in DiGraph.successors(self, n)
    925 try:
--> 926     return iter(self._succ[n])
    927 except KeyError as err:

KeyError: 'get_speed_0'

The above exception was the direct cause of the following exception:

NetworkXError                             Traceback (most recent call last)
Cell In[2], line 6
      4 first_child_label = next(iter(my_wf_dict["nodes"].keys()))
      5 my_wf_dict["label"] = first_child_label
----> 6 my_graph = get_knowledge_graph(my_wf_dict)
      7 validate_values(my_graph)

File ~/dev/pyiron/semantikon/semantikon/ontology.py:354, in get_knowledge_graph(wf_dict, include_t_box, include_a_box, hash_data, remove_data, extract_dataclasses, prefix, store_data, file_name, pmdco_uri)
    326 def get_knowledge_graph(
    327     wf_dict: dict,
    328     include_t_box: bool = True,
   (...)
    336     pmdco_uri: str = "https://w3id.org/pmd/co/3.0.0",
    337 ) -> Graph:
    338     """
    339     Generate RDF graph from a dictionary containing workflow information
    340 
   (...)
    352         (rdflib.Graph): graph containing workflow information
    353     """
--> 354     G = serialize_and_convert_to_networkx(wf_dict, hash_data=hash_data, prefix=prefix)
    355     graph = _get_bound_graph()
    356     graph += _import_pmdco(pmdco_uri=pmdco_uri)

File ~/dev/pyiron/semantikon/semantikon/ontology.py:1402, in serialize_and_convert_to_networkx(wf_dict, hash_data, prefix)
   1398         raise RuntimeError(
   1399             "Failed to hash workflow data - use only hashable inputs or set hash_data=False"
   1400         ) from e
   1401     for node, data in hashed_dict.items():
-> 1402         G.append_hash(node, data["hash"])
   1403 return G

File ~/dev/pyiron/semantikon/semantikon/ontology.py:160, in SemantikonDiGraph.append_hash(self, node, hash_value, label)
    157 while stack:
    158     current_node, current_hash, current_label = stack.pop()
--> 160     for child in self.successors(current_node):
    161         if self.nodes[child]["step"] == "node":
    162             continue

File ~/dev/miniforge3/envs/pyiron12/lib/python3.12/site-packages/networkx/classes/digraph.py:928, in DiGraph.successors(self, n)
    926     return iter(self._succ[n])
    927 except KeyError as err:
--> 928     raise NetworkXError(f"The node {n} is not in the digraph.") from err

NetworkXError: The node get_speed_0 is not in the digrap

```

The actual source of the problem is buried in `serialize_and_convert_to_networkx`:

https://github.com/pyiron/semantikon/blob/b31fa6c7e949b0313a29ad0f493262694bba794e/semantikon/ontology.py#L1370-L1371

Where the `hashed_dict` winds up with keys mismatched from the digraph labels because of that if-clause when the parent and child names match.

I see two paths forward:
1. Clarify and encode the restriction that names need to be globally and not just graph-layer unique[^1][^2]
2. Don't use that if clause

I've just pushed a branch doing (2). It works fine locally and is my preferred solution if it doesn't cause any side-effects I'm oblivious to.

[^1]: I looked at the README and both notebooks, and didn't find any naming conventions specified, but maybe there's stuff in a docstring somewhere.
[^2]: An obligatory comment about unstructured data (`wf_dict: dict`) being the devil.

liamhuber:
Otherwise, in the case where the key (child node name) and graph name (parent node name) are the same, the keys of the `hashed_dict` wind up out of sync with the labels available in the digraph.

Closes #421 

The problem is that I don't understand _why_ that if-clause is there to start with. So in the spirit of Chesterton's fence, this change should be scrutinized prior to merging.

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -1391,8 +1391,7 @@ def serialize_and_convert_to_networkx(
         try:
             hashed_dict = {}
             for key, value in get_hashed_node_dict(wf_dict).items():
-                if not key.startswith(G.name):
-                    key = G.name + "-" + key
+                key = G.name + "-" + key
                 hashed_dict[key.replace(".", "-")] = value
         except Exception as e:
             raise RuntimeError(


samwaseda:
There was a reason why I was doing it, but right now I cannot remember what it was....

samwaseda (approved):
Let's merge it for now. I see that the change was introduced in #375, and I see that I explicitly included this one step [here](https://github.com/pyiron/semantikon/commit/050f799aa47f3feaf84b9f3a34e9935a44cfb29c), but I cannot remember why I did it. At least your concern in #421 is valid, so it makes sense to address it here.

PR #424, opened 2026-04-13
Don't just `else` into `"outputs"`

liamhuber:
As we update formats, we might get an annoying bug where a name has changed but we don't notice because we `else` right into assuming we're dealing with `"outputs"`. Guard against this magic-string danger by making all clauses explicit.

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -948,13 +948,15 @@ def _nx_to_kg(G: SemantikonDiGraph, t_box: bool) -> Graph:
                 G=G,
                 t_box=t_box,
             )
-        else:
+        elif step == "outputs":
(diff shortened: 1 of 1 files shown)


liamhuber:
I don't know how to artificially manufacture an example that hits the edge case. This is basically just to give a cleaner exception for if the internal format gets incompletely updated...

liamhuber:
Nevermind, the `assert step in ["node", "inputs", "outputs"], f"Unknown step: {step}"` right above this means that `else`ing into `"outputs"` is safe unless the interpreter is run optimized `-O` _and_ a bug is introduced. There is not user facing data entry that can force us into an unknown step, so the odds of this being a problem are close to zero.

PR #426, opened 2026-04-14
Nested io

liamhuber:
There is a root paradigm is that the graph connectivity proceeds `input -> node -> output`. By including nested workflows such that parents pass their input to children, and populate their output from children, we get `input -> node + input -> child_input` and `node -> output + child_output -> output`, which breaks some of the validation. The place I'm running into this is in nested macros. Consider this simple case:

```python
import semantikon

def add_one(x):
    y = x + 1
    return y

@workflow
def add_two(a):
    b = add_one(a)
    c = add_one(b)
    return c

@workflow
def add_three(alpha):
    beta = add_two(alpha)
    gamma = add_two(beta)
    return gamma

wf_dict = add_three.get_semantikon_dict()
wf_dict

```

<details>
<summary>Which has a (IMO) reasonable and well formed dictionary representation</summary>

```
{'type': 'workflow',
 'label': 'add_three',
 'nodes': {'add_two_0': {'type': 'workflow',
   'label': 'add_two_0',
   'nodes': {'add_one_0': {'type': 'atomic',
     'function': <function __main__.add_one(x)>,
     'inputs': {'x': {}},
     'outputs': {'y': {}}},
    'add_one_1': {'type': 'atomic',
     'function': <function __main__.add_one(x)>,
     'inputs': {'x': {}},
     'outputs': {'y': {}}}},
   'edges': [('inputs.a', 'add_one_0.inputs.x'),
    ('add_one_0.outputs.y', 'add_one_1.inputs.x'),
    ('add_one_1.outputs.y', 'outputs.c')],
   'inputs': {'a': {}},
   'outputs': {'c': {}},
   'function': <function __main__.add_two(a)>},
  'add_two_1': {'type': 'workflow',
   'label': 'add_two_1',
   'nodes': {'add_one_0': {'type': 'atomic',
     'function': <function __main__.add_one(x)>,
     'inputs': {'x': {}},
     'outputs': {'y': {}}},
    'add_one_1': {'type': 'atomic',
     'function': <function __main__.add_one(x)>,
     'inputs': {'x': {}},
     'outputs': {'y': {}}}},
   'edges': [('inputs.a', 'add_one_0.inputs.x'),
    ('add_one_0.outputs.y', 'add_one_1.inputs.x'),
    ('add_one_1.outputs.y', 'outputs.c')],
   'inputs': {'a': {}},
   'outputs': {'c': {}},
   'function': <function __main__.add_two(a)>}},
 'edges': [('inputs.alpha', 'add_two_0.inputs.a'),
  ('add_two_0.outputs.c', 'add_two_1.inputs.a'),
  ('add_two_1.outputs.c', 'outputs.gamma')],
 'inputs': {'alpha': {}},
 'outputs': {'gamma': {}},
 'function': <function __main__.add_three(alpha)>}
```

</details>

Therefor, we'd like to be able to validate it:

```python
g = semantikon.get_knowledge_graph(wf_dict=wf_dict)
semantikon.validate_values(g)
```

Here this works fine, because I've modified the `_output_is_connected` and `_input_is_connected` checks to look (in a quacks-like-a-duck way) to see if we're looking at a port negotiating this sort of subgraph information flow.

On #425 it raises an error like:

```
----> 1 g = semantikon.get_knowledge_graph(wf_dict=wf_dict)

...

ValueError: Problematic candidates for add_three-add_two_1-inputs-a `_output_is_connected` check from among ['add_three-add_two_1', 'add_three-add_two_1-add_one_0-inputs-x']: {'add_three-add_two_1': {'step': 'node', 'type': 'workflow', 'label': 'add_two_1', 'function': {'module': '__main__', 'qualname': 'add_two', 'version': 'not_defined', 'hash': 'add_two:11dbfeca63783b3ec675a13c5d816b149b117fa49bf1f5104cb718fe7a497693', 'docstring': '', 'name': 'add_two', 'identifier': '__main__.add_two.not_defined'}, 'parent': 'add_three'}}
```

I.e. `add_three-add_two_1-inputs-a` is being swept up in an _output_ connection check, and fails because it has as its successor both its owning node (because `input -> node`) and the child input (because this is how information gets down into the subgraph).

At a minimum, I'll add tests based off this example to validate that the `SemantikonDigraph` generation handles nested topology. I also see that in #425 the "parent" info is explicitly tracked, so I might be able to use that to get the "is this negotiating subgraph data flow?" check to be more robust.

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -598,12 +598,12 @@ def _output_is_connected(io: str, G: SemantikonDiGraph) -> bool:
             return True
         return _output_is_connected(candidate[0], G)
     elif n_candidates == 2 and _is_macro_input(io, G, tuple(candidate)):
-        return (
-            _output_is_connected(candidate[0], G)
-            and _output_is_connected(candidate[1], G)
+        return _output_is_connected(candidate[0], G) and _output_is_connected(
+            candidate[1], G
         )
     else:
         return any(_output_is_connected(c, G) for c in candidate)
+
 
 def _is_macro_input(io: str, G: SemantikonDiGraph, candidates: tuple[str, str]):
     successor_types = {G.nodes[c]["step"] for c in candidates}
@@ -622,9 +622,8 @@ def _input_is_connected(io: str, G: SemantikonDiGraph) -> bool:
             return True
         return _input_is_connected(candidate[0], G)
     elif n_predecessors == 2 and _is_macro_output(io, G, tuple(candidate)):
-        return (
-            _input_is_connected(candidate[0], G)
-            and _input_is_connected(candidate[1], G)
+        return _input_is_connected(candidate[0], G) and _input_is_connected(
+            candidate[1], G
         )
     else:
         predecessor_steps = {c: G.nodes[c]["step"] for c in candidate}
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -590,20 +590,26 @@ def _wf_node_to_graph(
 
 def _output_is_connected(io: str, G: SemantikonDiGraph) -> bool:
     candidate = list(G.successors(io))
-    if len(candidate) == 1:
+    n_candidates = len(candidate)
+    if n_candidates == 0:
+        return False
+    elif n_candidates == 1:
         if G.nodes[candidate[0]]["step"] == "node":
             return True
         return _output_is_connected(candidate[0], G)
-    elif len(candidate) > 1:
-        if trouble := {
-            c: G.nodes[c] for c in candidate if G.nodes[c]["step"] == "node"
-        }:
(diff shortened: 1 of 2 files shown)


liamhuber:
@samwaseda, I don't see a way around this upstream assertion to get through to the exceptions not being covered here:

https://github.com/pyiron/semantikon/blob/3998a03afb3c0bfb5005d4b382ed7645e8fb35c2/semantikon/ontology.py#L115-L122

We could unit-test the private methods more directly, but I don't want to do a bunch on that front if we're anyhow going to [get an overhaul](https://github.com/pyiron/semantikon/pull/425#issue-4262700317)

samwaseda (approved):
I don't have a very strong opinion about this one. Fine by me

Issue #429, opened 2026-04-16
[bug] invalid parent-child data flow fails to invalidate


liamhuber:
```python
import rdflib

import semantikon
from semantikon.workflow import workflow

EX: rdflib.Namespace = rdflib.Namespace("http://example.org/")

def io_transformer(
    x: semantikon.u(int, uri=EX.Input),
) -> semantikon.u(int, uri=EX.Output):
    y = x
    return y


def input_sanitizer(
    x: int
) -> semantikon.u(int, uri=EX.Input):
    return x


def wrong_input_sanitizer(
    x: int
) -> semantikon.u(int, uri=EX.NotInput):
    return x

@workflow
def good_peer_edge(
    x_outer: int
) -> semantikon.u(int, uri=EX.Output):
    sanitized = input_sanitizer(x_outer)
    transformed = io_transformer(sanitized)
    return transformed

@workflow
def bad_peer_edge(
    x_outer: int
) -> semantikon.u(int, uri=EX.Output):
    wrongly_sanitized = wrong_input_sanitizer(x_outer)
    transformed = io_transformer(wrongly_sanitized)
    return transformed

@workflow
def good_parent_edge(
    x_outer: semantikon.u(int, uri=EX.Input)
) -> semantikon.u(int, uri=EX.Output):
    transformed = io_transformer(x_outer)
    return transformed


@workflow
def bad_parent_edge(
    x_outer: semantikon.u(int, uri=EX.NotInput)
) -> semantikon.u(int, uri=EX.Output):
    transformed = io_transformer(x_outer)
    return transformed

for wf in (
    good_peer_edge,
    bad_peer_edge,
    good_parent_edge,
    bad_parent_edge,
):
    d = wf.get_semantikon_dict()
    g = semantikon.get_knowledge_graph(d)
    v = semantikon.validate_values(g)
    print(wf.__name__, "valid?", v[0])
>>> good_peer_edge valid? True
>>> bad_peer_edge valid? False
>>> good_parent_edge valid? True
>>> bad_parent_edge valid? True

```

The `bad_parent_edge` should invalidate because of the mismatching URIs, as it does when the URI mismatch is between peer nodes, but it fails to do so. I guess this is related to #426 and the addition of parent nodes to the nx graph.

liamhuber:
I threw this (with dicitonary, graph triples, and the validation report printed out) along with the ontology module at Claude. It wrote a novel, but I think the key insight is this:

[...]

PR #431, opened 2026-04-16
Only emit bnode URIs from owned data

liamhuber:
For now, this PR is just to concretely demonstrate what I'm talking about in #429

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -338,7 +338,7 @@ def get_knowledge_graph(
     store_data: bool = False,
     file_name: str | None = None,
     pmdco_uri: str = "https://w3id.org/pmd/co/3.0.0",
-    enforce_subgraph_io_uri_alignment: bool = True,
+    enforce_subgraph_uri_alignment: bool = True,
 ) -> Graph:
     """
     Generate RDF graph from a dictionary containing workflow information
@@ -352,7 +352,7 @@ def get_knowledge_graph(
         extract_dataclasses (bool): if True, extract dataclass information into the graph
         prefix (str | None): prefix to use for the workflow namespace.
             If None, a hash-based prefix is generated.
-        enforce_subgraph_io_uri_alignment (bool): if True, parent workflow input
+        enforce_subgraph_uri_alignment (bool): if True, parent workflow input
             argument URIs must explicitly align with any child input URIs to which
             they are passed. Mechanistically, if True, when parsing the A-box only
             nodes whose name matches their associated data node will emit a uri
@@ -370,7 +370,7 @@ def get_knowledge_graph(
         graph += _nx_to_kg(
             G,
             t_box=False,
-            all_data_emit_uri=not enforce_subgraph_io_uri_alignment,
+            all_data_emit_uri=not enforce_subgraph_uri_alignment,
         )
     if extract_dataclasses:
         graph += extract_dataclass(
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -782,15 +782,22 @@ def _wf_input_to_graph(
             )
         if "restrictions" in data:
             g += _restrictions_to_triples(data["restrictions"], data_node=data_node)
-    else:
-        data_node = G.get_a_node(G._get_data_node(io=node_name))
+    else:  # ABox
+        data_node_name = G._get_data_node(io=node_name)
+        data_node = G.get_a_node(data_node_name)
         if not _input_is_connected(node_name, G):
-            if units is not None:
-                g.add((data_node, QUDT.hasUnit, _units_to_uri(units)))
-            if "uri" in data:
-                bnode = URIRef(str(data_node) + "_uri")
-                g.add((bnode, RDF.type, data["uri"]))
-                g.add((data_node, SNS.specifies_value_of, bnode))
+            # Only emit URI/unit info if this node OWNS the data node.
+            # If the data node traces back to a parent input, that parent is the
+            # canonical holder; emitting here would collide with the parent's
+            # emission on the shared bnode and mask uri/unit mismatches.
+            owns_data_node = data_node_name == f"{node_name}_data"
+            if owns_data_node:
+                if units is not None:
+                    g.add((data_node, QUDT.hasUnit, _units_to_uri(units)))
+                if "uri" in data:
+                    bnode = URIRef(str(data_node) + "_uri")
+                    g.add((bnode, RDF.type, data["uri"]))
+                    g.add((data_node, SNS.specifies_value_of, bnode))
     g += _wf_io_to_graph(
         node_name=node_name,
         data=data,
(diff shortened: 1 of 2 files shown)


liamhuber:
I'm not going to add demonstrating tests, because it sounds in #429 like this is an anti-solution. But just in case that changes, this a foundation for writing such tests:

```python
import rdflib

import semantikon

EX: rdflib.Namespace = rdflib.Namespace("http://example.org/")

def io_transformer(
    x: semantikon.u(int, uri=EX.Input),
) -> semantikon.u(int, uri=EX.Output):
    y = x
    return y


def input_sanitizer(
    x: int
) -> semantikon.u(int, uri=EX.Input):
    return x


def wrong_input_sanitizer(
    x: int
) -> semantikon.u(int, uri=EX.NotInput):
    return x

@workflow
def good_peer_edge(
    x_outer: int
) -> semantikon.u(int, uri=EX.Output):
    sanitized = input_sanitizer(x_outer)
    transformed = io_transformer(sanitized)
    return transformed

@workflow
def bad_peer_edge(
    x_outer: int
) -> semantikon.u(int, uri=EX.Output):
    wrongly_sanitized = wrong_input_sanitizer(x_outer)
    transformed = io_transformer(wrongly_sanitized)
    return transformed

@workflow
def good_parent_edge(
    x_outer: semantikon.u(int, uri=EX.Input)
) -> semantikon.u(int, uri=EX.Output):
    transformed = io_transformer(x_outer)
    return transformed


@workflow
def bad_parent_edge(
    x_outer: semantikon.u(int, uri=EX.NotInput)
) -> semantikon.u(int, uri=EX.Output):
    transformed = io_transformer(x_outer)
    return transformed

for wf in (
    good_peer_edge,
    bad_peer_edge,
    good_parent_edge,
    bad_parent_edge,
):
    d = wf.get_semantikon_dict()
    g = semantikon.get_knowledge_graph(d)
    v = semantikon.validate_values(g)
    print(wf.__name__, "valid?", v[0])
```

samwaseda:
I don't oppose this PR, but this doesn't fully solve the problem, because if there are more layers, only the outermost type will be checked against the inner types.

liamhuber:
> I don't oppose this PR, but this doesn't fully solve the problem, because if there are more layers, only the outermost type will be checked against the inner types.

Yes, I think you're right that it doesn't work for more depths. For me, the incompleteness of the solution makes me oppose the PR.

My understanding is still limited, but the better my understanding gets the more I think the ports sharing data bnodes is problematic.

liamhuber:
I went to concretely build a counter example for posterity, but the thing I _thought_ would fail actually passes fine.

```python
@workflow
def child_wf(
    x_child: semantikon.u(int, uri=EX.Input)
) -> semantikon.u(int, uri=EX.Output):
    out = io_transformer(x=x_child)
    return out

@workflow
def middle_wf(
    x_mid: semantikon.u(int, uri=EX.Input)
) -> semantikon.u(int, uri=EX.Output):
    out = child_wf(x_child=x_mid)
    return out

@workflow
def parent_wf(
    x_outer: semantikon.u(int, uri=EX.Input)
) -> semantikon.u(int, uri=EX.Output):
    mid = middle_wf(x_mid=x_outer)
    return mid
```

This validates fine, but if any one of `x_outer`/`x_mid`/`x_child` is changed to use `uri=EX.NotInput`, it fails to validate.

@samwaseda, was there some other layered example you had in mind?

I'm going to re-open in the meantime, because I think it's interesting and it gets a handful of old `pyiron_workflow` tests written for 0.0.22 to work out-of-the-box. But, I think you _ought_ to oppose it 😂 This hard-wires closed-world validation, and I thought you wanted open-world validation. If I have to chose one, I want closed-world, but ideally I'd want the `validate_values` call to take a flag letting me choose.

liamhuber:
Behaviour is also nice for a bifurcating case:

```python
@workflow
def child_wf(
    x_child: semantikon.u(int, uri=EX.Input)
) -> semantikon.u(int, uri=EX.Output):
    out = io_transformer(x=x_child)
    return out

def anti_io_transformer(
    x: semantikon.u(int, uri=EX.NotInput),
) -> semantikon.u(int, uri=EX.Output):
    y = x
    return y

@workflow
def anti_child_wf(
    x_child: semantikon.u(int, uri=EX.NotInput)
) -> semantikon.u(int, uri=EX.Output):
    out = anti_io_transformer(x=x_child)
    return out

@workflow
def bifurcating_wf(
    x_outer: semantikon.u(int, uri=EX.Input)
) -> tuple[
    semantikon.u(int, uri=EX.Output),
    semantikon.u(int, uri=EX.Output),
]:
    good = child_wf(x_child=x_outer)
    bad = anti_child_wf(x_child=x_outer)
    return good, bad
```

Which fails to validate (since the anti-child is not getting what it wants). I got Claude to look at the resulting full set of triples to help my understand the mechanism by which this patch is making it fail:

> Grepping the dump confirms: I see `allValuesFrom EX.NotInput` and `allValuesFrom EX.Input` both attached to the same `x_outer_data` TBox class via different bnodes. Both become `sh:class` constraints on the `specifies_value_of` target. The target bnode is typed only `EX.NotInput` — so the `sh:class EX.Input` constraint fails, validation fails.
>
> This confirms the patch catches the mismatch even in the bifurcating case. The bnode is singular, the constraints are conjunctive, and a mismatch on any edge of the bifurcation surfaces as a SHACL failure against that one bnode.

To the best of my still limited understanding, this is indeed the behaviour I want when validating a (possibly nested) annotated workflow.

samwaseda:
Ah it's true :D ok then let's merge this one and we're done?

samwaseda (approved):

samwaseda:
nevertheless we should also merge #432 because we would probably need the logic to validate other things

liamhuber:
> Ah it's true :D ok then let's merge this one and we're done?

Almost. This actually breaks an existing test: the meal example where we currently show that annotating something as `EX.Water` doesn't fail until we make `EX.Water` and `EX.Meal` disjoint. With this PR we force closed reasoning such that when it doesn't find `EX.Meal` it already complains.

Right now my plan would be:
- [ ] Create a new `"closed"/"open"` reasoning enum and expose it in the `validate_values` function, then propagate it through to the changes here so we can toggle old/new behaviour
  - @samwaseda what should the default be, closed or open? I guess I mildly prefer closed as default.
- [ ] Use open reasoning for the test that currently fails behaves as it does on `main:HEAD`
- [ ] Add an inverted version of the failing test so we see that after adding `subclassOf`, closed-world validation does pass
- [ ] Add the multilayer and bifurcating examples from this thread to the test suite

samwaseda:
I'm fine with making closed default

liamhuber:
Super. I'll work on the plan proposed above and ping you for re-review when I'm done. I expect this to be done later today or tomorrow in my time zone.

liamhuber:
I need clarity on #439 before being able to proceed with this plan.

liamhuber:
Unfortunately, this breaks `test_inheritance_of_derives_from`, where we have an un-annotated parent wrapping an annotated child. From __, the all-in-one-place test code is:

```python
from typing import Annotated

import rdflib
import semantikon

EX = rdflib.Namespace("http://example.org/")

def f_triples(
    a: Annotated[float, {"uri": EX.Something}],
    b: Annotated[float, {"triples": ("self", EX.relatedTo, "inputs.a")}],
) -> Annotated[
    float,
    {
        "triples": ((EX.hasSomeRelation, "inputs.b")),
        "derived_from": "inputs.a",
        "uri": EX.SomethingElse,
    },
]:
    return a


@workflow
def wf_triples(a, b):
    a = f_triples(a, b)
    return a

d = wf_triples.get_semantikon_dict()
g = semantikon.get_knowledge_graph(d, enforce_subgraph_uri_alignment=True)
# semantikon.visualize_recipe(g)
onto._inherit_properties(g)
uri_node = list(g.subjects(rdflib.RDF.type, EX.Something))
data_node = list(g.subjects(onto.SNS.specifies_value_of, uri_node))
uri_node, data_node
```

Where I've explictly written out the new toggle, which when True forces restricts the emission of `_uri` nodes to when the node owns the data node specifying the URI. The problem is, that when you turn this on `g.subjects(rdflib.RDF.type, EX.Something)` is totally empty.

I can imagine that this is somehow related to the data node and owning node never aligning since the parent workflow is not itself annotated. The problem is that when I look at the `semantikon.visualize_recipe(g)` it looks identical, so I'm clearly failing to understand something quite critical.

Not adequately passing inferred information back upstream seems quite sketchy to me though, so for now I just close this.

PR #432, opened 2026-04-17
Translate `owl:disjointWith` into shacl shape

samwaseda:
```python
import rdflib

import semantikon

EX: rdflib.Namespace = rdflib.Namespace("http://example.org/")


def io_transformer(
    x: semantikon.u(int, uri=EX.Input),
) -> semantikon.u(int, uri=EX.Output):
    y = x
    return y

@workflow
def bad_parent_edge(
    x_outer: semantikon.u(int, uri=EX.NotInput)
) -> semantikon.u(int, uri=EX.Output):
    transformed = io_transformer(x_outer)
    return transformed

kg = semantikon.get_knowledge_graph(bad_parent_edge.get_semantikon_dict())
kg.add((EX.Input, rdflib.OWL.disjointWith, EX.NotInput))
kg.add((EX.Input, rdflib.RDF.type, rdflib.OWL.Class))
kg.add((EX.NotInput, rdflib.RDF.type, rdflib.OWL.Class))

semantikon.validate_values(kg)
```

This now gives `False` because `ex:NotInput` and `ex:Input` are disjoint and no object can provide them both at the same time.

### DIFF
--- a/semantikon/ontology.py
+++ b/semantikon/ontology.py
@@ -1549,10 +1549,13 @@ class _OWLToSHACLConverter:
         """
         self.owl_graph = owl_graph
         self.excluded_nodes = excluded_nodes
-        self.shacl_graph = _get_bound_graph()
-        self.node_shapes: dict[URIRef, BNode] = {}
-        self.shacl_graph.bind("sh", str(SH))
-        self.shacl_graph.bind("sns", str(BASE))
+
+    @staticmethod
+    def _get_bound_graph():
+        shacl_graph = _get_bound_graph()
+        shacl_graph.bind("sh", str(SH))
+        shacl_graph.bind("sns", str(BASE))
+        return shacl_graph
 
     def _iter_supported_restrictions(self):
         """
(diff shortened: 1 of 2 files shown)


liamhuber:
I won't have time to review the code before my meeting starts, but I am absolutely in favour of the direction. `disjointWith` is, IMO, a very intuitive predicate, so being able to easily leverage it for URI mutual exclusivity is nice. So I'm in favour.

However, this still doesn't address my core concern that validation should pass when "everything's right" rather than when "nothing's explicitly wrong". That's not this PR's fault though, so I'll continue that discussion there.

samwaseda:
> However, this still doesn't address my core concern that validation should pass when "everything's right" rather than when "nothing's explicitly wrong". That's not this PR's fault though, so I'll continue that discussion there.

But that's the typical conflict between open and closed world assumptions, so I would argue it is just a question of preference and there's nothing wrong about it

samwaseda:
@copilot Open another PR to account for other OWL properties

liamhuber:
> > However, this still doesn't address my core concern that validation should pass when "everything's right" rather than when "nothing's explicitly wrong". That's not this PR's fault though, so I'll continue that discussion there.
> 
> But that's the typical conflict between open and closed world assumptions, so I would argue it is just a question of preference and there's nothing wrong about it

Right, I agree, but my point is this PR _does not_ close #429. I like the PR, I just don't think it resolves the issue.

liamhuber (approved):
Code and test lgtm! Some form of the test needs to wind up in the notebook at some point so the capability is advertised, but this doesn't need to be now.

liamhuber:
I can't remove the closes connection... Here there's no interactivity, and on the issue itself I can un-check the box but then it immediately and automatically re-checks

PR #435, opened 2026-04-20
Update numpy to 2.4.3

jan-janssen:


samwaseda (approved):
I approved it because you asked for a review, but I don't think there's something I should minutely scrutinize, right?

jan-janssen:
@samwaseda ideally we would have done these before the 1.0 release. Now we can either release a 1.0.1 version or we simply patch the conda package to include these changes. The first is more clear, the later a bit quicker, which do you prefer?

samwaseda:
Fine by me with both options - you can choose

jan-janssen:
I proposed the corresponding patch in https://github.com/conda-forge/semantikon-feedstock/pull/44

PR #442, opened 2026-04-21
Test specificity directionality

liamhuber:
For URI annotations.

No changes here, everything's working exactly as I hoped it would. I'd just like to keep it that way and I don't see this represented in the existing tests.

### DIFF
--- a/tests/unit/test_ontology.py
+++ b/tests/unit/test_ontology.py
@@ -54,7 +54,6 @@ def add_three(alpha):
     return gamma
 
 
-# def upstream_node(a) -> meta.u(object, uri=EX.Upstream):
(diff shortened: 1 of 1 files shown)


tests/unit/test_ontology.py
```diff
@@ -54,6 +54,24 @@ def add_three(alpha):
     return gamma
 
 
+# def upstream_node(a) -> meta.u(object, uri=EX.Upstream):
```
liamhuber:
```suggestion
```

liamhuber (commented):
Just trying to commit a suggestion from mobile

samwaseda (approved):

PR #455, opened 2026-06-01
Flowrep 2 Semantikon

liamhuber:
I'm playing around in `pyiron_workflow` trying to directly validate flowrep objects, and I discovered that the `flowrep.NodeData` -> `semantikon` `dict` converter was straight-up missing the ontological metadata. At the end of the day, once I added those in (in `_port_dict` for the `u` stuff, and higher up the call chain for the `@metadata` function-level stuff), I discovered that I could remove `workflow.to_semantikon_workflow_dict` since it no longer added anything extra. This gave a cascade of unused code such that the `workflow` module is now just a very thin wrapper setting up the decorator architecture. The cost of this was some very minor additions to `flowrep_dict`, so overall I see this as a big deduplication win.

Along the road to figuring all this out, I discovered (via `deepcopy` interaction) that `SemantikonURI` turns into a plain old `URIRef` on de-serialization roundtripping and fixed that.

This doesn't change any existing behaviour, and just extends the API a bit so it's another minor bump.


samwaseda:
I separated this PR into #456 and #457 because the changes appear orthogonal to each other

liamhuber:
Superseded by #456 and #457

PR #456, opened 2026-06-01
Buff up SemantikonURI serialization

samwaseda:
I discovered this when we deepcopied the workflow dictionary, that deepcopying SemantikonURI objects gave back plain old URIRef objects. The deepcopy is gone, so it's no longer strictly necessary in this PR, but IMO it's a nice improvement.

### DIFF
--- a/semantikon/metadata.py
+++ b/semantikon/metadata.py
@@ -159,6 +159,18 @@ def meta(
     return decorator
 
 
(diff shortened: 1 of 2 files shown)


samwaseda:
Ok since I opened this PR I cannot approve it, but nice catch! Thanks for correcting it!

PR #457, opened 2026-06-01
Flowrep annotation metadata

samwaseda:
This pull request refactors and enhances the workflow serialization logic in the `semantikon` package, focusing on improving type annotation handling and simplifying the API for converting workflow nodes to dictionaries. The changes rename and standardize the core node-to-dictionary function, add richer type metadata extraction, and remove legacy code that is now redundant due to these improvements. Tests are updated accordingly to use the new interfaces and to cover the new metadata features.

**API and Function Renaming/Refactoring:**

- The main function for converting workflow nodes to dictionaries is renamed from `node_data_to_dict` to `nodedata2dict` throughout the codebase, and all imports and usages are updated accordingly. This simplifies the API and makes naming more consistent. [[1]](diffhunk://#diff-346d0a41b374d14cc6521d3828108fd29de3e1572e792b0aa72ec41a2e5a025dR48-R53) [[2]](diffhunk://#diff-346d0a41b374d14cc6521d3828108fd29de3e1572e792b0aa72ec41a2e5a025dL126-R130) [[3]](diffhunk://#diff-1e0c27527336f363a42c5fbb6dba97e9c5444bf2cd8bf894a7c8932ad5e5ce72L253-R32) [[4]](diffhunk://#diff-32f6a1d53a04c7f53db9f59b13547cd8e0805356279538a2b729ca765b0daa1cL159-R180) [[5]](diffhunk://#diff-32f6a1d53a04c7f53db9f59b13547cd8e0805356279538a2b729ca765b0daa1cL176-R196) [[6]](diffhunk://#diff-32f6a1d53a04c7f53db9f59b13547cd8e0805356279538a2b729ca765b0daa1cL192-R220) [[7]](diffhunk://#diff-32f6a1d53a04c7f53db9f59b13547cd8e0805356279538a2b729ca765b0daa1cL216-R231) [[8]](diffhunk://#diff-32f6a1d53a04c7f53db9f59b13547cd8e0805356279538a2b729ca765b0daa1cL227-R254) [[9]](diffhunk://#diff-32f6a1d53a04c7f53db9f59b13547cd8e0805356279538a2b729ca765b0daa1cL249-R264) [[10]](diffhunk://#diff-32f6a1d53a04c7f53db9f59b13547cd8e0805356279538a2b729ca765b0daa1cL260-R281) [[11]](diffhunk://#diff-32f6a1d53a04c7f53db9f59b13547cd8e0805356279538a2b729ca765b0daa1cL277-R292)

[...]

PR #459, opened 2026-06-09
API tools accept flowrep objects

liamhuber:
Namely `get_knowledge_graph` and `validate_values` (which is also extended to take a (presumably semantikon-formatted) dict. Per discussion in #457, this lets us remove `nodedata2dict` from the API, and makes the API more approachable for flowrep users. All super straightforward type coercion.

As a final step, I changed the `with_io/function` defaults in `nodedata2dict` from `False` to `True`. @samwaseda, my only outstanding question is whether we can further streamline things by getting rid of these flag arguments altogether and force-setting the `True` state; the `...=False` flags only ever appear in the test suite to explicitly verify the falsey behaviour, so it looks to me like we wouldn't get any downstream side effects. Of course, we can always leave that for a separate PR -- the key thing here is the API improvements.


semantikon/flowrep_dict.py
```diff
@@ -53,8 +53,8 @@
 def nodedata2dict(
     node: frs.NodeData,
     *,
-    with_io: bool = False,
-    with_function: bool = False,
+    with_io: bool = True,
```
samwaseda: I actually don't think `with_io = False` makes sense in semantikon. From my point of view, we can hard code `with_io = True` below

liamhuber: Will do!

liamhuber: Done. There was one "push on the explicit failure when we set it to False" test I had to remove, but it was an otherwise trivial change 🚀

samwaseda:
Actually `with_function` can also be hard-coded to be `True` I think.

liamhuber:
> Actually `with_function` can also be hard-coded to be `True` I think.

👍 Done!

samwaseda (approved):
Yep looks good to me!

PR #460, opened 2026-06-10
Fix ruff import sorting in test_analysis.py

samwaseda:
**Summary**

Fixes failing CI job `pyiron / ruff-sort-imports` caused by incorrect import ordering in `tests/unit/test_analysis.py`.

**Related Issue(s)**

**Backward Compatibility**

No backward compatibility concerns — only test file import ordering was changed.

**Detailed Description**

The `ruff` import sorter (isort rules) flagged two issues in `tests/unit/test_analysis.py`:

1. In `test_function_request_pydantic_validation`: imports were ordered as `FunctionRequest, Annotation` — corrected to alphabetical order `Annotation, FunctionRequest`.
2. In `test_extract_annotations_from_graph`: `from rdflib import URIRef` appeared after `from semantikon.analysis import extract_annotations_from_graph` — corrected to place the third-party (`rdflib`) import before the first-party (`semantikon`) import, with a blank line separating the two groups per isort convention.


semantikon/analysis.py
```diff
@@ -7,12 +7,62 @@
...
+    name: str
+    type_: str | None = None
+    label: str | None = None
+    position: int | None = None
+    default: Any = None
+    uri: str | None = None
+    restrictions: dict[str, Any] | None = None
+
+    class Config:
+        extra = "allow"
```
liamhuber, line 30: What will the relationship be between this class and

https://github.com/pyiron/semantikon/blob/3cf8fafe27b55a3c3b30a470e312f0d7d98d7816/semantikon/datastructure.py#L74-L81

? Is `TypeMetadata` being deprecated, or are these supposed to live in parallel to serve fundamentally different roles? There are differences between the structure of fields, e.g. `restrictions` is `dict[str, Any]` here and `tuple[str, str] | tuple[tuple[str, str,], ...] | tuple[ tuple[ tuple[str, str], ...], ...]` there, name mismatches (e.g. "name" vs "label", and field set differences ((units, shape, derived_from), (type, position, default)).

samwaseda:
@liamhuber I thought about this PR once again and I think it's total nonsense to host the functionality in `semantikon`. In order to make the functionality complete, I'll commit a few changes more, but then I'll close this PR.

liamhuber:
> @liamhuber I thought about this PR once again and I think it's total nonsense to host the functionality in `semantikon`. In order to make the functionality complete, I'll commit a few changes more, but then I'll close this PR.

No worries, I fully trust your judgement either way. I'm fully on board that we want to collect and store nice metadata, I think we're all on the same page there, but I don't yet have any super clear views on the best path(s) to accomplish it.

Just to keep it on your radar, the `flowrep`-based implementation of `pyiron_workflow` has a space to [associate metadata with the retrospective result of running a workflow](https://github.com/pyiron/pyiron_workflow/blob/7d87c216b330c220c587121ed2480d3b4abc719e/pyiron_workflow/_wfms/execution.py#L38-L61). That's tangential to what this PR was doing, but it's meta-data-related and I just want it floating around in your head.

semantikon/analysis.py
samwaseda: The data class comes from the node store, and it's not really meant for an actual schema. But as I already said below, I don't think this should be hosted in semantikon, so it probably doesn't matter anyway

XzzX:
I am totally fine with moving everything here into the sim-atlas toolkit. I will draft this and link it here.

PR #471, opened 2026-06-23
feat: Base ontology parsing on flowrep data class and not wf_dict

samwaseda:
well, wf_dict used to be called flowrep, but apparently things have changed a lot since then.

For the backward compatibility, wf_dict should also still be supported. It's just transformed back to flowrep data classes.


samwaseda:
@liamhuber Since this changes nothing for the user interface, I merge it without asking for your review. I don't think it's worth your time to read the changes.

PR #472, opened 2026-06-24
feat: Use flowrep format for request values

samwaseda:
Same as #471 - `request_values` takes now flowrep objects and returns `DagData` instead of `dict`.

Do not merge this one until we are ready for a minor bump, because the output type of `request_values` is not `dict` anymore.


samwaseda:
I'm not going to fix the notebook problem until we are about to merge this one because it often leads to a conflict

samwaseda:
@copilot Update the notebook according to the changes here

samwaseda:
@copilot It's still failing. Fix it.

PR #473, opened 2026-06-28
Translate knowledge graph to flowrep

samwaseda:
**Summary**

Adds backward transformation from a Semantikon knowledge graph to flowrep, and now also includes flowrep-style alias names for the converter API: `kg2recipe` and `kg2data`.

**Related Issue(s)**

N/A

**Backward Compatibility**

No breaking changes. Existing `knowledge_graph_to_flowrep_recipe` and `knowledge_graph_to_flowrep_data` remain available; the new names are additive aliases.

**Implementation Notes**

Implemented and exposed graph-to-flowrep conversion through the public API, added the `kg2recipe`/`kg2data` aliases for naming consistency with flowrep converters, and updated unit coverage to exercise the alias-based path.

**Example**

```python
import flowrep as fr
import semantikon
import networkx as nx
from semantikon.ontology import SNS
import rdflib

def add_one(a):
    """some docstring"""
    return a + 1

def add(x, y):
    return x + y

@fr.workflow
def add_more(x):
    y = add_one(x)
    z = add(x, y)
    return z

@fr.workflow
def my_workflow(x=1, y=2):
    z = add_one(x)
    q = add_more(z)
    result = add(q, y)
    return result

graph = semantikon.get_knowledge_graph(my_workflow.flowrep_recipe)
print(fr.wfms.run_recipe(semantikon.kg2recipe(graph)).output_ports == fr.wfms.run_recipe(my_workflow.flowrep_recipe).output_ports
)
```

Output: `True`


liamhuber:
Something like `knowledge2recipe` would be more in line with the flowrep converter names, if you like it

samwaseda:
@copilot Address @liamhuber's comment

samwaseda:
@liamhuber I think I'd rather use `kg` and not `knowledge` because somehow `knowledge` sounds like something other than knowledge graph

liamhuber:
> @liamhuber I think I'd rather use `kg` and not `knowledge` because somehow `knowledge` sounds like something other than knowledge graph

👍💯

PR #481, opened 2026-07-06
Cwl windows ci

liamhuber:


liamhuber:
Ok @samwaseda, here's the patches to break apart the env for windows tests. The steps are not necessarily clear a-priori, but it's worth skimming the changes because IMO it's at least very clear retrospectively.

The hiccup is the failing mypy. Honestly, I think there's nothing wrong here -- these are the exact tests that fail when I run mypy locally. What I don't understand (and have not understood for some time) is how on `main` mypy passes. I've just been quietly assuming that `main` is using an older mypy and not catching as much. In https://github.com/pyiron/semantikon/pull/482 I just exclude these so they can be fixed at a later date.

liamhuber:
> I've just been quietly assuming that `main` is using an older mypy and not catching as much.

The mypy versions (and accompany modules mypy-extensions, typing-extensions) are all the exact same, so it's not just that mypy is older. I have no idea what the actual difference is.

samwaseda:
Is it possible that mypy is not recognizing the yml files correctly and cannot know what imported modules do?

liamhuber:
> Is it possible that mypy is not recognizing the yml files correctly and cannot know what imported modules do?

I'm not sure what yml files you mean. The failure here is correct, I also get it on my local machine. The only mystery to me is how `pyiron/actions-4.0.8` _didn't_ fail for these modules. I think it has something to do with `pip install mypy` and `python -m pip install mypy`, which is the main difference I can see between 4.0.8 and 4.2.1. The newer one installs it inside the context of the overall conda env, and if you look in the recent logs you can see the `ENV:` differs.

At any rate, it's not a huge deal. In #482 I just exclude these files from the analysis manually. We can then add them back to the mypy coverage in separate PR(s).

samwaseda:
oh sorry I didn't read your previous message carefully (as always XD). I guess we should fix this first no?

liamhuber:
I don't feel strongly. As far as I can tell these mypy complaints are legitimate. So an end state where mypy complains is desirable. That's what we get here, then in the stacked PR we explicitly tell mypy to be quiet (until we have time to actually fix the underlying type problems). I don't see a solution significantly better than that.

PR #482, opened 2026-07-06
Exclude mypy-failing files

liamhuber:
Just brute-force to get the green checkmark back. Obviously in the long run we want these to be covered too.


samwaseda:
I close this one because the issue has been resolved in #483

PR #491, opened 2026-07-14
feat: Use pydantic class for SemantikonDiGraph

samwaseda:
Initially `SemantikonDiGraph` was just a local representation, but now that
1. loading of workflows from KG and
2. parsing of CWL

runs over it, I decided to make it a bit more solid by introducing pydantic classes

### DIFF
--- a/semantikon/flowrep_to_networkx.py
+++ b/semantikon/flowrep_to_networkx.py
@@ -27,7 +27,6 @@ class NodeData(pydantic.BaseModel):
     model_config = pydantic.ConfigDict(frozen=True)
 
     type: str
-    step: str = "node"
     identifier: str | None = None
     label: str | None = None
     function: dict | None = None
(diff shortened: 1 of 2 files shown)


liamhuber:
You can use an enum for `type` for even more robustness/clarity. In fact, I think doing so might remove the need for manually validating that bit.

What I think would be really cool is if you could declare it as a subset of the corresponding flowrep enum; then the terms "workflow" and "atomic" would actually get a hard link. Idk if exclusionary subclassing of enum like that is possible though.

samwaseda:
it looks like I cannot simply use data classes without changing the structure of `SemantikonDiGraph` because the data has to be serialized before it's passed to `SemantikonDiGraph` and when the metadata is given by data classes, instead of the data classes themselves the serialized version is passed to `SemantikonDiGraph`. So I keep this PR simple and do the change to data classes once I have a better idea how to do it.

PR #492, opened 2026-07-14
feat: Include function in the workflow hashing algorithm

samwaseda:
I realized that the underlying python function is currently not included in the hashing algorithm for the workflows. Now I included the hash of the function or the function identifier. To be honest since it's semantikon-dependent I might not like it so much...

### DIFF
--- a/semantikon/flowrep_to_networkx.py
+++ b/semantikon/flowrep_to_networkx.py
@@ -414,10 +414,9 @@ class _HashGraph:
         for node in G.nodes:
             attrs = {}
             if "function" in G.nodes[node]:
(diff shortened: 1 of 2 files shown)


liamhuber (commented):
I'm not technically aware enough to understand how `G.nodes[node]["function"]` is a type of object for which `.get("hash")` is a meaningful thing to grab, but I'm happy to accept that it is. My hope is it's more or less built of the information you find in `pyiron_snippets.versions.VersionInfo`.

So, with that caveat on my technical knowledge shortcomings, yes, totally agree, function info should absolutely make it into the hash and it looks like it's propagating to the hash in the test value just fine 👍

PR #499, opened 2026-07-23
Handle functionless workflows

liamhuber:
@samwaseda, 1.4.0 regressively breaks the parsing of workflows without an underlying python reference function with these lines:

https://github.com/pyiron/semantikon/blob/528fbff1af198046081382accd64a9887b4d419d/semantikon/flowrep_dict.py#L174-L177

Now, we don't currently have a way to attach any ontological information to those, so any KG/onto information on them is going to be really boring, and maybe this was on purpose. On the other hand, their worst crime is just being boring, so maybe this regression was an accident.

I got Claude to put together a little patch. AFAIK I don't actually need this functionality at present, but if the fallout radius here was bigger than intended we can mend the lost functionality here. Up to you, the patch came almost for free from getting the AI to try and debug anything that could be done on the pwf side so I thought I might as well pass it on.

### DIFF
--- a/semantikon/flowrep_dict.py
+++ b/semantikon/flowrep_dict.py
@@ -173,9 +173,19 @@ def _normalize_output_label(label: str, recipe_outputs: list[str]) -> str:
 
 def _dict_to_workflow_recipe(node: dict[str, Any]) -> fr.schemas.WorkflowRecipe:
     function = node.get("function")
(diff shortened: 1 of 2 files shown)


samwaseda (approved):
ah good point. Thanks for fixing it!

PR #500, opened 2026-07-27
Test with numpy version 1.26.3

jan-janssen:

PR #501, opened 2026-07-27
fix ruff problems

samwaseda:
which first appeared in #500


samwaseda:
@copilot address the problems raised in ruff-check

PR #508, opened 2026-07-30
Parse flowrep constants

liamhuber:
Parses flowrep constant nodes into both the A-box and T-box in analogy to how existing nodes treat function references. In contrast to the `rdflib.URIRef` function reference, the new constant reference uses `rdflib.Literal`. Then this is propagated through so the resulting recipes still round-trip, and things still visualize pleasantly.

**On main:**

```python
import flowrep as fr
import semantikon

@fr.atomic
def two():
    return 2

@fr.workflow
def wf_main(x):
    scalar_two = two()
    doubled = fr.std.mul(scalar_two, x)
    return doubled

kg_main = semantikon.get_knowledge_graph(wf_main.flowrep_recipe)
semantikon.visualize_recipe(kg_main)
```

atomic_output

**Here:**

```python
@fr.workflow
def wf(x):
    doubled = fr.std.mul(2, x)
    return doubled

kg = semantikon.get_knowledge_graph(wf.flowrep_recipe)
semantikon.visualize_recipe(kg)
```

You can see in the visualization that new constant node behaves just like the existing nodes, but instead of "concretizes" for a function reference, it "has_value" of a literal

constant_output

The resulting KG round-trips back to the original recipe

```python
>>> round_trip_recipe = semantikon.kg2recipe(kg)
>>> print(round_trip_recipe.model_dump_json(indent=2))
{
  "type": "workflow",
  "inputs": [
    "x"
  ],
  "outputs": [
    "doubled"
  ],
  "description": null,
  "nodes": {
    "mul_0": {
      "type": "atomic",
      "inputs": [
        "a",
        "b"
      ],
      "outputs": [
        "product"
      ],
      "description": null,
      "reference": {
        "info": {
          "module": "flowrep.std",
          "qualname": "mul",
          "version": null
        },
        "inputs_with_defaults": [],
        "restricted_input_kinds": {
          "a": "POSITIONAL_ONLY",
          "b": "POSITIONAL_ONLY"
        }
      }
    },
    "constant_0": {
      "type": "constant",
      "inputs": [],
      "outputs": [
        "constant"
      ],
      "description": null,
      "constant": 2
    }
  },
  "input_edges": {
    "mul_0.b": "x"
  },
  "edges": {
    "mul_0.a": "constant_0.constant"
  },
  "output_edges": {
    "doubled": "mul_0.product"
  },
  "reference": {
    "info": {
      "module": "__main__",
      "qualname": "wf",
      "version": null
    },
    "inputs_with_defaults": [],
[...]

PR #511, opened 2026-08-04
Parse constants to KG

### LINKED ISSUE #507: Graphs with `fr.schemas.ConstantRecipe` fail to parse
```python

import flowrep as fr
import semantikon

@fr.workflow
def wf(x):
    doubled = fr.std.mul(2, x)
    return doubled

semantikon.get_knowledge_graph(wf.flowrep_recipe)
# semantikon.validate_values(wf.flowrep_recipe)  # same problem

```

<details><summary>`AttributeError: 'ConstantRecipe' object has no attribute 'reference'`</summary>

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 semantikon.get_knowledge_graph(wf.flowrep_recipe)

File ~/dev/pyiron/semantikon/semantikon/ontology.py:291, in get_knowledge_graph(wf_dict, include_t_box, include_a_box, hash_data, remove_data, extract_dataclasses, prefix, store_data, file_name, pmdco_uri)
    261 def get_knowledge_graph(
    262     wf_dict: dict | fr.schemas.DagData | fr.schemas.WorkflowRecipe,
    263     include_t_box: bool = True,
   (...)
    271     pmdco_uri: str = "https://w3id.org/pmd/co/3.0.0",
    272 ) -> Graph:
    273     """
    274     Generate RDF graph from workflow information
    275 
   (...)
    289         (rdflib.Graph): graph containing workflow information
    290     """
--> 291     G = serialize_and_convert_to_networkx(wf_dict, hash_data=hash_data, prefix=prefix)
    292     return _get_knowledge_graph_from_digraph(
    293         G,
    294         include_t_box=include_t_box,
   (...)
    300         pmdco_uri=pmdco_uri,
    301     )

File ~/dev/pyiron/semantikon/semantikon/flowrep_to_networkx.py:507, in serialize_and_convert_to_networkx(workflow, hash_data, prefix)
    500 if not isinstance(workflow, fr.schemas.DagData):
    501     raise TypeError(
    502         f"Invalid workflow type. Expected dict, flowrep"
    503         f" {fr.schemas.DagData.__name__!r}, or flowrep"
    504         f" {fr.schemas.WorkflowRecipe.__name__!r}, but got {type(workflow)}."
    505     )
--> 507 G = _workflow_to_networkx(workflow, prefix=prefix)
    508 if hash_data:
    509     try:

File ~/dev/pyiron/semantikon/semantikon/flowrep_to_networkx.py:411, in _workflow_to_networkx(workflow, prefix)
    405             src_port = _output_port_label(source.port, child_outputs)
    406             G.add_edge(
    407                 f"{node_name}-{source.node}-outputs-{src_port}",
    408                 f"{node_name}-outputs-{target_port}",
    409             )
--> 411 _add_node(workflow, root_label, workflow_label=root_label)
    412 return G

File ~/dev/pyiron/semantikon/semantikon/flowrep_to_networkx.py:374, in _workflow_to_networkx.<locals>._add_node(node_data, node_name, parent_name, workflow_label)
    372 for child_label, child in node_data.nodes.items():
    373     child_name = f"{node_name}-{child_label}"
--> 374     _add_node(
    375         child,
    376         child_name,
    377         parent_name=node_name,
    378         workflow_label=(
    379             child_label if isinstance(child, fr.schemas.DagData) else None
    380         ),
    381     )
    383 child_recipes = recipe.nodes
    384 for target, source in recipe.input_edges.items():

File ~/dev/pyiron/semantikon/semantikon/flowrep_to_networkx.py:309, in _workflow_to_networkx.<locals>._add_node(node_data, node_name, parent_name, workflow_label)
    307     if workflow_label is not None:
    308         metadata["label"] = workflow_label
--> 309     if node_data.recipe.reference is not None:
    310         function = retrieve.import_from_string(
    311             node_data.recipe.reference.info.fully_qualified_name
    312         )
    314 if function is not None:

File ~/dev/miniforge3/envs/pyiron12/lib/python3.12/site-packages/pydantic/main.py:1026, in BaseModel.__getattr__(self, item)
   1023     return super().__getattribute__(item)  # Raises AttributeError if appropriate
   1024 else:
   1025     # this is the current error
-> 1026     raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')

AttributeError: 'ConstantRecipe' object has no attribute 'reference'

```

</details>

Since [workflows without a reference are fine](https://github.com/pyiron/semantikon/pull/499), I think this represents a simple plumbing problem rather than any conceptual concern.

samwaseda:
Closes #507 

Comments will follow


samwaseda:
I should have stacked from #508...

samwaseda:
@copilot Resolve the mypy issue

samwaseda:
@copilot Fix the mypy problem

samwaseda:
For me to remember: In this PR I only burned tons of tokens to get only the parser from KG to SemantikonDiGraph out, but then the algorithm is also extremely fragile, as instead of properly using the ontological classes, it looks at the node names. An acute reminder that AI does not necessarily do what I want

liamhuber:
Coverage and mypy still failing...

semantikon/flowrep_to_networkx.py
```diff
@@ -467,10 +469,23 @@ def _get_hashed_node_dict_from_graph(G: SemantikonDiGraph) -> dict[str, dict[str
...
+            for inp in G.successors(output_node):
+                G.nodes[inp]["constant_value"] = const_value
+    G.remove_nodes_from(to_delete)
+
+
 def serialize_and_convert_to_networkx(
     workflow: dict | fr.schemas.DagData | fr.schemas.WorkflowRecipe,
     hash_data: bool = True,
     prefix: str | None = None,
+    remove_constant_nodes: bool = True,
```
liamhuber: I just checked for usages of `serialize_and_convert_to_networkx` -- this flag is never used. Under what conditions do we _not_ want to remove constant nodes when converting to networkx in this new paradigm? Naively, it looks like this flag is dead weight and can be removed.

liamhuber: Hmm, although it's not exposed to `get_knowledge_graph`, if I force this flag to `False` and visualize the resulting KG it looks exactly like #508 except that the constant node doesn't expose a `has_value` field.

It is more difficult for me to understand the intent and constraints of what is an "allowable" KG when both these representations exist with a simple public flag switch, and both generate a KG with out error and pass a simple `semantikon.validate_values`...

liamhuber: Ok, but the resulting KG fails to round trip back into a recipe. I am increasingly convinced that this should not be exposed as a flag, but should forcibly always happen.

semantikon/kg_to_flowrep.py
```diff
@@ -564,6 +571,114 @@ def _build_workflow_graph(graph: Graph) -> nx.DiGraph:
...
+
+    for node, data in tuple(G.nodes.data()):
+        if data.get("step") != "inputs" or "constant_value" not in data:
+            continue
+
+        # Extract the parent node and argument name
+        # Node format: parent-inputs-arg
+        parts = node.rsplit("-", 2)
+        assert len(parts) == 3, f"Unexpected input node format: {node}"
+        assert parts[1] == "inputs", f"Expected 'inputs' in node name: {node}"
```
liamhuber, line 636: This is tangential to this PR, but all these assertions really smell like we need more formalized structure. The node format is `parent-inputs-arg` -- where does this come from? Can we make new functions/(pydantic?) classes that produce and consume parent/IO/arg in a symmetric and structured way?

Right now there is some place that is producing this string pattern, but it's opaque to this consumer location. If we ever go back and change that production pattern, it will be annoying to track down and update locations like this where we consume it.

semantikon/kg_to_flowrep.py
```diff
@@ -564,6 +571,114 @@ def _build_workflow_graph(graph: Graph) -> nx.DiGraph:
...
+            parent_data.get("step") == "node"
+        ), f"Parent node {parent_node} is not a workflow node for {node}"
+
+        # Create constant node name using a unique counter
+        assert (
+            "parent" in parent_data
+        ), f"Parent node {parent_node} is missing 'parent' attribute for {node}"
+        parent_prefix = parent_data["parent"] + "-"
+
+        for constant_index in range(1000):
```
liamhuber: Why 1000? It is a stretch, but I can imagine some workflow sometime having 1000 nodes. Why not just `range(sys.maxsize)`?

If there is no strict digit limitation inside the other tools,

```suggestion
        for constant_index in itertools.count():
```

even more directly emphasizes the intent here, and then there is simply no upper bound to worry about.

semantikon/kg_to_flowrep.py
```diff
@@ -564,6 +571,114 @@ def _build_workflow_graph(graph: Graph) -> nx.DiGraph:
...
+        ), f"Parent node {parent_node} not found for {node}"
+        assert (
+            parent_data.get("step") == "node"
+        ), f"Parent node {parent_node} is not a workflow node for {node}"
+
+        # Create constant node name using a unique counter
+        assert (
+            "parent" in parent_data
+        ), f"Parent node {parent_node} is missing 'parent' attribute for {node}"
+        parent_prefix = parent_data["parent"] + "-"
```
liamhuber, line 651: Similar to the above comment. Here I'd imagine something like instantiating some (pydantic?) class `parent_data = ParentData(G.nodes.get(parent_node))` that validates the data like the assertions and has a nice property field like `.prefix: str`. The benefit is then when the underlying data gets produced something like `pd.serialize()` for some `pd: ParentData`. Then the expectations casting to and from the different formats can be coupled together and more easily kept in sync.

Maybe there are complications that make this not so straightforward, and is certainly outside the scope of this PR, but it is worth thinking about as this sort of assertion system is fragile and hard to maintain.

semantikon/kg_to_flowrep.py
```diff
@@ -564,6 +571,114 @@ def _build_workflow_graph(graph: Graph) -> nx.DiGraph:
...
+
+        for constant_index in range(1000):
+            const_node_name = f"{parent_prefix}constant_{constant_index}"
```
liamhuber: I'm trying to get a handle on what the multitude of `"constant"` strings actually mean and are actually maintained by.

In this case, I think changing the label might formally break round-tripping because the resulting node labels differ, but it won't _meaningfully_ break it because the labels are just labels and don't impact topology. Still, I think we can at least more clearly indicate the rationale behind this hard coded value.

```suggestion
            const_node_name = f"{parent_prefix}{fr.schemas.ConstantRecipe.std_label}_{constant_index}"
```

semantikon/kg_to_flowrep.py
```diff
@@ -564,6 +571,114 @@ def _build_workflow_graph(graph: Graph) -> nx.DiGraph:
...
+            if const_node_name not in G:
+                break
+
+        const_output_name = f"{const_node_name}-outputs-constant"
+
+        # Create the constant node
+        const_node_attrs = TNode(
+            type="constant",
```
liamhuber, line 664: I'm trying to get a handle on what the multitude of `"constant"` strings actually mean and are actually maintained by.

It would be nice if the `TNode` `type` took a custom `StrEnum` that could take the values currently validated in the `__post_init__`. In fact, it might be nice if `TNode` and `TIO` were pydantic models instead of dataclasses to handle this validation. 

This is outside scope here.

semantikon/kg_to_flowrep.py
```diff
@@ -564,6 +571,114 @@ def _build_workflow_graph(graph: Graph) -> nx.DiGraph:
...
+            parent=parent_data.get("parent"),
+        )
+        G.add_node(const_node_name, **const_node_attrs.to_attrs())
+
+        # Create the constant output node
+        const_output_attrs = TOutput(
+            arg="constant",
```
liamhuber: I'm trying to get a handle on what the multitude of `"constant"` strings actually mean and are actually maintained by.

Looking over at `_networkx_to_dict`, for `"constant"`-type nx nodes, we don't even use the `arg`, only the `value` to reconstruct a `flowrep` node. So while I initially thought this _had_ to be `fr.schemas.ConstantRecipe.std_label` because otherwise [flowrep would refuse to validate](https://github.com/pyiron/flowrep/blob/4f65e6ff3c4203146ebd55885b34f9c3063293f7/src/flowrep/prospective/constant_recipe.py#L87-L91), it never actually gets passed to flowrep and could be anything we want.

_However_, it _is_ constrained because when I manually change it to `"constantz"`, the workflow recipe actually fails to validate with 

```
ValidationError: 1 validation error for WorkflowRecipe
  Value error, Invalid edge source port: constant_0.constantz [type=value_error, input_value={'inputs': ['x'], 'output...}), 'description': None}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/value_error
```

Aha, probably because over in `_networkx_to_dict` we force-instantiate a `ConstantRecipe` with the output `"constant"`, but in the next line we add nodes using `**const_output_attrs.to_attrs()`. My guess is that the bad `arg` here trickles out to bad edges when we convert back to flowrep. So I can't concretely prove it, but I think this MUST be


```suggestion
            arg=fr.schemas.ConstantRecipe.std_label,
```

semantikon/kg_to_flowrep.py
```diff
@@ -674,8 +774,31 @@ def kg2recipe(
...
+        workflow_name (str | URIRef | None): Optional root workflow identifier. Can be:
+            - String: Local identifier label of the workflow
+            - URIRef: Direct URIRef to the workflow
+            - None: Required only when the graph contains a single root workflow.
+
+    Returns:
+        fr.schemas.WorkflowRecipe: Flowrep workflow recipe.
+    """
+    selected_workflow = _kg2digraph(graph, workflow_name=workflow_name)
+    return _networkx_to_dict(selected_workflow)
```
liamhuber, line 806: Out of scope for this PR, but `_networkx_to_dict` has become a very misleading name

liamhuber (commented):
The Good (/enough):
- Functionality appears bang-on, and the tests (especially round-tripping) indicate this PR does everything needed 🚀 
- In all honesty, I still find it easier to understand a KG that more closely represents the flowrep graph by having an explicit node for constants. With that said, the setup here where constants hard-code a value specification is merely _less_ clear to me, not _unclear_, and I can live with it 👍 

The Bad:
- I'm quite convinced `remove_constant_nodes` should be removed from the arguments of `serialize_and_convert_to_networkx` and _always_ be done
- Some of the new `"constant"` strings should or need to be locked to `fr.schemas.ConstantRecipe.std_label`

The Ugly (and out of scope):
- There are too-damned-many magic strings in this project. It makes it incredibly hard to understand the constraints on any given value, or find what other pieces of the code base they are related to.

semantikon/flowrep_to_networkx.py
samwaseda: I just removed it. I needed it at the beginning for debugging reasons, but then also I wasn't sure whether the public NetworkX graph should look closer to flowrep or KG, but I guess by postponing the decision I just ended up somewhere in between...

semantikon/kg_to_flowrep.py
samwaseda: That's a very good point. I'll look into it right after this PR.

semantikon/kg_to_flowrep.py
samwaseda: I committed this one, but the node name anyway does not really matter, right?

liamhuber: Indeed, here it's safe to stray. But you chose `"constant"` because it's a "constant" node, which is the same reason `fr.schemas.ConstantRecipe.std_label = "constant"`. I just think it's nice if the same reasoning can exploit the same choice, which it can easily do since `std_label` is a class var.

PR #512, opened 2026-08-12
chore: Remove the case that child node name does not start with paren…

samwaseda:
…t node name


samwaseda:
If the tests do not fail, it was my fault to not have properly checked the code written by AI.

samwaseda:
And they didn't... I don't know if I should be happy or sad

PR #513, opened 2026-08-17
Use data classes for SemantikonDiGraph

samwaseda:
Following [Liam's comment](https://github.com/pyiron/semantikon/pull/511#pullrequestreview-4900364233) I replaced all the strings in `SemantikonDiGraph` by data classes. In particular, there are the following ones:

```python
@dataclass(frozen=True, slots=True)
class Node:
    name: str
    parent: Node | None = None


@dataclass(frozen=True, slots=True)
class IO(ABC):  # this is separated into `Input` and `Output`
    node: str
    arg: str
```

This entirely changes the content of `SemantikonDiGraph`, but it is not given in the API, so I consider it as a patch.

### DIFF
--- a/semantikon/analysis.py
+++ b/semantikon/analysis.py
@@ -16,7 +16,7 @@ from rdflib.query import ResultRow
 
 from semantikon.converter import to_identifier
 from semantikon.flowrep_dict import dict_to_nodedata
-from semantikon.flowrep_to_networkx import IO, Input, Node, Output
+from semantikon.flowrep_to_networkx import Input, Node, Output
 from semantikon.ontology import SNS, serialize_and_convert_to_networkx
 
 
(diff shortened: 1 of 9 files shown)


liamhuber:
I think it's not necessary but may be nice to align language with flowrep -- ie port instead of arg, and maybe owner instead of parent

liamhuber:
Ah and Label instead of name?

liamhuber:
SDG and the flowrep graphs are not strictly the same thing, so these are only suggestions. The only one I'd push hard on is arg/port. It belong to IO, so for input "arg" confuses with "kwarg" and for Output "arg" doesn't make sense at all.

samwaseda:
@copilot Update semantikon/cwl.py and tests/unit/test_cwl.py

samwaseda:
@copilot Resolve mypy problems

samwaseda:
There are a lot more modifications I would like to make related to the data class introduction, but since this PR is getting really long, and there is anyway no change to the API, I merge this one and open new PRs.

PR #520, opened 2026-08-24
fix: use prefix for the name if it is not defined

samwaseda:


samwaseda:
@copilot Write a unit test for the changes made here.

PR #528, opened 2026-09-11
Update numpy to 2.5.3 and bagofholding to 0.1.13

jan-janssen:
Just like `bagofholding` in https://github.com/pyiron/bagofholding/pull/160 this drops testing for Python 3.11, which is required as `numpy=2.5.3` only supports Python 3.12 to 3.15:
https://github.com/numpy/numpy/releases


jan-janssen:
```
          The following packages are incompatible
          ├─ bagofholding =0.1.12 * is installable and it requires
          │  └─ numpy >=1.26.4,<2.5.0 * with the potential options
          │     ├─ numpy 2.5.0rc1 would require
          │     │  └─ _numpy_rc =* *, which does not exist (perhaps a missing channel);
          │     └─ numpy [2.1.0|2.1.1|...|2.4.6], which can be installed;
          └─ numpy =2.5.3 * is not installable because it conflicts with any installable versions previously reported.
```

jan-janssen:
@samwaseda Can you merge this and release a new `semantikon` version?

